PR #10, opened 2019-07-20
Roadmap proposal

kaelig:
Here's a proposal for driving this effort over the next few months.

Once we have a good idea of the format (see #1), we should start defining plain colors (and maybe support opacity too), as having a universal specification for colors would have the highest impact in design systems, and reaching agreement shouldn't be too complex compared to other design properties such as spacing, time, breakpoints, and grids.

Both a simplified visual breakdown and detailed breakdown are available in this pull request.

timeline

Questions/comments/nods are welcome in the comments of this pull request!

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ We believe that a common way to share design tokens will unlock efficiency oppor
 
 The DTWG was founded in June 2019. By the end of 2019, our aim is to have a standard working across at least 3 design or prototyping tools.
 
-![timeline](https://user-images.githubusercontent.com/85783/61584564-ba912400-aafe-11e9-919a-c059c50a084e.png)
+![timeline](https://user-images.githubusercontent.com/85783/61584753-422c6200-ab02-11e9-8066-e594c8eca946.png)
 
 ### June 2019
 
@@ -54,6 +54,11 @@ The DTWG was founded in June 2019. By the end of 2019, our aim is to have a stan
 - Open source tools (Theo, Style Dictionary, and maybe others like System UI…) build pre-releases to support the working draft
 
 ### November 2019
+
+- Community / public testing
+- Recommendation: format (and maybe theming too, depending on how earlier discussions go)
+
+### December 2019
 
 - Recommendation: colors level 1
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ We believe that a common way to share design tokens will unlock efficiency oppor
 
 The DTWG was founded in June 2019. By the end of 2019, our aim is to have a standard working across at least 3 design or prototyping tools.
 
+![timeline](https://user-images.githubusercontent.com/85783/61584564-ba912400-aafe-11e9-919a-c059c50a084e.png)
+
 ## Who sits on the DTWG
 
 The working group is composed of UX professionals, developers, and representants of design tooling vendors.
--- a/README.md
+++ b/README.md
@@ -20,15 +20,19 @@ We believe that a common way to share design tokens will unlock efficiency oppor
 
 ## Roadmap
 
-The DTWG was founded in June 2019. By the end of 2019, our aim is to have a standard working across at least 3 design or prototyping tools.
+The DTWG was founded in June 2019. By the end of 2019, our aim is to have a basic implementation working across:
+
+- at least 3 design or prototyping tools
+- additional design token management tools such as Theo, Style Dictionary, Invision DSM, and Diez
+- various open source community-driven projects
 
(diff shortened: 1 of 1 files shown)


kaelig:
Very good point, thank you for the review @c1rrus.

I amended the part you refer to in https://github.com/design-tokens/working-group/pull/10/commits/a23452e99fb1669987e52f9896fd610d91bbf880.

README.md
```diff
@@ -53,3 +105,18 @@ We are currently reaching out to various design software vendors to get a repres
 We acknowledge that the format specification is only part of an ecosystem, supporting methods and practices that relate to scaling design tokens:
 
 [![a tweet by @jina, posted on the 14th of November 2018 – "Design Tokens are a methodology. IMHO, saying “design tokens are just variables” is like saying “responsive design is just media queries”. It’s a technology-agnostic architecture and process for scaling design across multiple platforms and devices, including native, and more."](https://user-images.githubusercontent.com/85783/59956860-eeb5ee00-9447-11e9-8ea3-3bb6c158bc1a.png)](https://twitter.com/jina/status/1062808011301965825)
+
+---
+
+## Glossary
```
dbanksdesign: This is awesome to have especially for those (like me) unfamiliar with these concepts! Above it mentions a 'candidate draft', should we include that in the glossary too?

README.md
```diff
@@ -20,7 +20,59 @@ We believe that a common way to share design tokens will unlock efficiency oppor
...
+
+- Recommendation: colors level 1
+
+### 2020
+
+Issue recommendations on additional aspects of design tokens:
+
+- Dimensions
+- Colors level 2 (such as gradients, etc)
+- Grid
```
dbanksdesign: + Motion (timing and easing)?
+ Groups / compound tokens (like borders & typography)?

README.md
```diff
@@ -20,7 +20,59 @@ We believe that a common way to share design tokens will unlock efficiency oppor
...
+- Editor’s draft: colors level 1
+
+### September 2019
+
+- Working draft: colors level 1
+- Collect feedback on the colors working draft
+
+### October 2019
+
+- Candidate Recommendation: colors level 1
```
dbanksdesign: Are we missing 'candidate draft's for format and theming?

mathieudutour:
Do we actually need a different token format for colors, typography, etc.? I’m not convinced that’s true and believe that we can find a format generic enough to be used by any tokens, which just a “value” changing. But this roadmap implies the opposite.

kaelig:
>Do we actually need a different token format for colors, typography, etc.?

My assumption was that we would need to agree on the type and shape of values. For example, how to ensure that all tools can import/export colors: should we use `hsla`, `rgba`, `hex8`, component (like for Sketch palettes)? What should the format for gradients be?

Does this make sense, or do you feel like there are other ways to achieve the same interoperability?

c1rrus:
What @kaelig just described was my interpretation too.

Considering @dbanksdesign earlier comment about the goals, it might be worth explicitly including at least the format.

I had _assumed_ that colors implied the file format as well, but it will be clearer if we say so explicitly.

Just to be clear, my (quite possibly incorrect) interpretation of these terms is as follows:

* **format** refers to the spec of the overall file format, within which platform agnostic design tokens are stored. I'd expect this to define things like:
    * Is it JSON, YAML, XML or something competely bespoke?
    * What's the structure? E.g. do we have a flat array of design tokens, or do we have nested objects and/or arrays of them?
    * What props can tokens have? name and value seem obvious, but there may be more for various kinds of metadata. I think we're leaning towards some notion of type (since we want to begin with colours). Also, there was some talk about a place for apps to store custom data that travels with the tokens.
    * How do we do aliases (or "references"). I.e. what's the syntax for one token to reference another for its value (and perhaps also other props).
    * Any other format-level features. E.g. we might want to allow some kind of file import mechanism, so that authors have the option of splitting up their design tokens across many files for easier maintenance.
* **colors** then refers to the details of tokens of type color. As already mentioned, this is probably just making a decision on how to encode color values, whether or not to allow an alpha channel a perhaps what kind of precision to use (24bit or something higher)?

If others have interpreted the intent of the WG's current goals differently, then please say so, so that we can discuss and all decide what exactly we want to be working on.

jina:
somehow missed this. watching the repo now. 😄 is this roadmap still on track or do the dates need to be updated?

nagueva:
Since we have an open pull request, can I ask if this roadmap is still valid? What we can expect as the next achievements?

kaelig:
Hey @nagueva – this PR is out of date, we'll rework it soon – closing for now!

PR #14, opened 2019-08-13
Add Superposition to list of companies

Kilian:
I just announced the design tokens app I've been working on, so I can add it to the list now :)


kaelig:
Thank you Kilian!

PR #28, opened 2019-11-21
add Alaska Air to the projects represented group

blackfalcon:
This PR will add Alaska Airlines to the list of represented groups.


kaelig:
Welcome! I just flew with you last week ✈️

PR #46, opened 2021-04-23
chore: update company name + website info for Knapsack

sghoweri:
Updates the contrib list to use our updated company name + website!


kaelig:
Thank you!

PR #51, opened 2021-09-07
Renamed vendor attribute to extensions

c1rrus:
Renames the `vendor` design token attribute to `extensions` to better reflect that it can be used by anyone (individuals, design system teams and tool vendors).


kevinmpowell (approved):
👍 Looks good.

kaelig (approved):

PR #52, opened 2021-09-07
Add composite types section

kaelig:
Copy-pasted from the working copy!

Example titles could be improved, please feel free to suggest alternatives.

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -6,7 +6,6 @@
 
 The types described in the previous chapter are all for singular values (a color, a dimension, etc.). However, it can often be useful to group related values as a single token so that they can be used or referenced as a single unit. A typical example of this is a “typography style” as found in many design tools, which is a combination of the font name, weight, style, and color.
 
-
 Other examples might be:
 
 - **color pairs**, e.g. a combination of foreground and background color
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -10,8 +10,8 @@ AND: A combination of the above two examples. Mutually exclusive composites. (Gr
 
 Other examples might be:
 
-- **color pairs**, e.g. a combo of foreground and background color
-- **shadows**, e.g. a combination of color, blur (a dimension value), x & y offsets (also dimension values),, and opacity (a number value)
+- **color pairs**, e.g. a combination of foreground and background color
+- **shadows**, e.g. a combination of color, blur (a dimension value), x & y offsets (also dimension values), and opacity (a number value)
 - **border styles**, e.g. a combination of color, style & thickness
 - **color schemes**
 - **text styles**, e.g. a combination of font and other font properties
@@ -117,7 +117,7 @@ Using the color pair type as an example, an export tool like Style Dictionary co
 
 </aside>
 
-….might be exported to Sass like this:
+… might be exported to Sass like this:
 
 <aside class="example" title="Sass output with fallback">
 
@@ -128,13 +128,13 @@ $my-token-background: #ffffff;
 
 </aside>
 
-In a similar vein, a GUI tool can "pluck" out the individual values of a composite token and use them as it would normally.
+In a similar vein, a <abbr title="Graphical User Interface">GUI</abbr> tool can "pluck" out the individual values of a composite token and use them as it would normally.
 
 E.g. a design tool like Figma might not have the concept of a color pair, so it can't do anything special with tokens of that type. However, that doesn't prevent it from displaying the foreground and background colors of those tokens alongside any plain color tokens in a color picker.
 
 ### Appending custom type definitions
 
-While a tool that imports a token file containing user-defined type definitions that it doesn't understand might not be able to do anything special with them, there's nothing stopping tools that export or modify token files from adding their own type definitions to the file alongside tokens of that type.
+A tool might not understand certain user-defined type definitions in a token file and therefore can't do anything special with them. However, that tool could export or modify token files by adding their own type definitions.
 
 For example, imagine Sketch wanted to export all layer styles from a Sketch document to a token file. The Sketch app could have a built-in type definition that corresponds to layer styles. Perhaps something like this:
 
@@ -166,7 +166,7 @@ For example, imagine Sketch wanted to export all layer styles from a Sketch docu
 
 </aside>
 
-It could then just save out that type definition into the token file it exports. Other tools could then make use of that same type definition and do useful things with tokens of that type.
+It could then save out that type definition into the token file it exports. Other tools could then make use of that same type definition and do useful things with tokens of that type.
 
 ### Custom configuration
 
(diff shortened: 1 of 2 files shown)


technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+
+</aside>
+
+In a similar vein, a GUI tool can "pluck" out the individual values of a composite token and use them as it would normally.
+
+E.g. a design tool like Figma might not have the concept of a color pair, so it can't do anything special with tokens of that type. However, that doesn't prevent it from displaying the foreground and background colors of those tokens alongside any plain color tokens in a color picker.
+
+### Appending custom type definitions
+
+While a tool that imports a token file containing user-defined type definitions that it doesn't understand might not be able to do anything special with them, there's nothing stopping tools that export or modify token files from adding their own type definitions to the file alongside tokens of that type.
```
jina, line 137: This is a very long sentence. Can we break it up for better readability?

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+</aside>
+
+It could then just save out that type definition into the token file it exports. Other tools could then make use of that same type definition and do useful things with tokens of that type.
```
jina, line 169: let's avoid using words like "Just" in our docs for better inclusivity.

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+}
+```
+
+</aside>
+
+The downside is of course that teams need to set up and maintain these configurations themselves.
+
+### GUIs
+
+Since composite types are ultimately composed of the core design token types, a design tool could automatically display an appropriate UI for creating, editing, or previewing them.
```
jina, line 209: same note here for `abbr` for UI.

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+
+For instance, the color pair example above consists of 2 colors, one with the key “foreground” and the other with the key “background”. A design tool could therefore display two color picker widgets using those keys as the respective labels.
```
jina, line 211: should we use code wraps instead of quotes for the keys (foreground & background)?

jina (changes requested):
Left a few comments. In general, I wonder if we want to use tools like Hemingway or Grammarly to make sure our documentation is inclusive and more understandable. Some of the sentences we have in our docs in general are very hard to grok. :)

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+```
+
+</aside>
+
+## Benefits of composite types over groups
+
+At first glance, groups and composite types might look very similar. However, they are intended to solve different problems and therefore have some important differences:
+
+- Groups are for arbitrarily grouping tokens for the purposes of naming and/or organization.
+  - They impose no rules or restrictions on how many tokens or nested groups you put within them, what they are called, or what the types of the tokens within should be. As such, tools MUST NOT try to infer any special meaning or typing of tokens based on a group they happen to be in.
```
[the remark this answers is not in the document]
kaelig: It's customary to follow this RFC https://www.ietf.org/rfc/rfc2119.txt – I agree it's definitely "shouty", but it seems like other W3C specs either use italics or simply lowercase with no formatting. It's probably easier to just use lowercase.

I suggest we do it all at once across all files once all sections are brought back from the Google Doc.

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
...
+<aside class="example" title="Sass output with fallback">
+
+```css
+$my-token-foreground: #333333;
+$my-token-background: #ffffff;
+```
+
+</aside>
+
+In a similar vein, a GUI tool can "pluck" out the individual values of a composite token and use them as it would normally.
```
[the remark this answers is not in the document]
kaelig: I agree – seems like it needs to be done once, and then ReSpec will automatically match all instances: https://respec.org/docs/#abbreviations

kaelig:
I agree with the comments from @jina – format module editors: I'll let you take this PR from here!

technical-reports/format/composite-types.md
dbanksdesign: I tried updating it, how does it sound now?

> A tool might not understand certain user-defined type definitions in a token file and therefore can't do anything special with them. However, that tool could export or modify token files by adding their own type definitions.

dbanksdesign:
I addressed @jina's comments

technical-reports/format/composite-types.md
kaelig: This can be re-edited later, but I feel like there's an opportunity for clear technical guidance here.

What exactly should happen when a tool doesn't understand user-defined types? Is there a cause and effect between this and the second sentence in the paragraph, or can they be two separate independent statements?

technical-reports/format/composite-types.md
```diff
@@ -206,11 +206,11 @@ The downside is of course that teams need to set up and maintain these configura
 
 ### GUIs
 
-Since composite types are ultimately composed of the core design token types, a design tool could automatically display an appropriate UI for creating, editing, or previewing them.
+Since composite types are ultimately composed of the core design token types, a design tool could automatically display an appropriate <abbr title="User Interface">UI</abbr> for creating, editing, or previewing them.
 
-For instance, the color pair example above consists of 2 colors, one with the key “foreground” and the other with the key “background”. A design tool could therefore display two color picker widgets using those keys as the respective labels.
+For instance, the color pair example above consists of 2 colors, one with the key <code>foreground</code> and the other with the key <code>background</code>. A design tool could therefore display two color picker widgets using those keys as the respective labels.
```
kaelig, line 211: Note that you can also use markdown back ticks

technical-reports/format/composite-types.md
c1rrus: In hindsight, this "Appending custom type definitions" section might be redundant. As it stands, composite tokens require a custom type definition. Therefore, any tool wanting to output composite tokens will necessarily need to embed the corresponding type definition in the token file.

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,220 @@
+# Composite types
+
+<p class="ednote" title="This section will be revisited">
+  Note: this proposal may not be in scope for the first version of the specification and requires further discussion.
+</p>
+
+The types described in the previous chapter are all for singular values (a color, a dimension, etc.). However, it can often be useful to group related values as a single token so that they can be used or referenced as a single unit. A typical example of this is a “typography style” as found in many design tools, which is a combination of the font name, weight, style, and color.
+
+AND: A combination of the above two examples. Mutually exclusive composites. (Grid styles)
```
c1rrus, line 9: I think this line should be removed - it's part of a comment in the original Google doc, but doesn't make sense on its own like this.

```suggestion
```

c1rrus (changes requested):
Thanks for migrating this chapter over. Just added a minor, suggested change.

technical-reports/format/composite-types.md
jina: yeah my review feedback is more on the grammar of a long run-on sentence — I'll let y'all figure out if you want to add or remove this. :)

technical-reports/format/composite-types.md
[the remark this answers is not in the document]
jina: can we track that then so this feedback doesn't get lost? It doesn't appear that all caps are required. just the language part.

technical-reports/format/composite-types.md
```diff
@@ -0,0 +1,214 @@
+# Composite types
+
+<div class="issue" data-number="54" title="Should composites be part of the MVP specification?">
+
+If so, which composites should be included initially?
+
+- Would composites allow for better integration with design tools?
```
kaelig, line 7: Can this be voiced as an open question, rather than a binary question?

technical-reports/format/composite-types.md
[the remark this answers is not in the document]
jina: created an issue from this

technical-reports/format/composite-types.md
jina: created issues for the other parts of this review. this is the last item standing before I can close the review. Do we know how we want to go forward with this?

PR #57, opened 2021-09-08
Prettier pre-commit hook

dbanksdesign:
I got annoyed that a pull request was failing a linting check, but not telling me why it was failing. Instead, what repositories generally do is have a pre-commit hook that runs prettier on files that were changed in the commit and writes the prettier changes. This way the contributor won't have to remember to run prettier and fix the issue by hand. Open to suggestions though!


.vscode/settings.json
```diff
@@ -1,4 +1,6 @@
 {
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
   "editor.formatOnSave": true,
-  "editor.formatOnPaste": true
+  "editor.formatOnPaste": true,
+  "files.insertFinalNewline": true
```
kaelig, line 5: This should be taken care of by editorconfig (which I forgot to add to the list of recommended extensions, sorry!)?

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
kaelig, line 4: Not sure it makes a difference but should this say `yarn run lint-staged` since we're using yarn on this repo?

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
```
kaelig, line 2: @c1rrus can you check if this works in Windows (without the Linux subsystem)?

kaelig (commented):
Thank you! A few questions - perhaps one blocker if it only works on Unix systems and not Windows.

.husky/pre-commit
c1rrus: I think you might be able to just to `lint-staged` without `npx` or `yarn run` in front of it.

.husky/pre-commit
c1rrus: I can, but it might be a while before I find the time to set it all up (got quite a lot going on at work and at home right now). I therefore suggest not holding up this PR. If I find issues when I eventually get around to trying it, I can always open a follow-up PR to address them then.

.husky/pre-commit
kaelig: Forget what I said, looks like Yarn + Husky have Windows compat issues. https://typicode.github.io/husky/#/?id=yarn-on-windows

.vscode/settings.json
kaelig: (I added editorconfig to the list of recommended extensions, do we still need this?)

.husky/pre-commit
kaelig: According to their docs, it works in Windows (as long as one has installed the appropriate tooling).

dbanksdesign:
Closing this in favor of #70

PR #71, opened 2021-09-14
feat: add specificity to token types. closes #58

### LINKED ISSUE #58: [Format] Token value - string, number, array, object?
JSON supports string, number, array and object as data types. Which of these do we want to support for token _values_.

String:
```
{
  colorTextPrimary: {
    value: '#000000'
  }
}
```

Number:
```
{
  durationDefault: {
    value: 10
  }
}
```

Array:
```
{
  breakpoints: {
    value: ['320', '640', 1080]
  }
}  
```

Object:
```
{
  textHeadingLevel1: {
    value: {
      fontSize: 32
      fontWeight: 700
      lineHeight: 1.5
    }
  }
}
```

I suspect we'll need to support all four data types.

kevinmpowell:
@c1rrus mind taking a look at this one when you get a chance? It resolves #58

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -23,9 +23,18 @@ The example above therefore defines 1 design token with the following properties
 
 Name and value are both **required**.
 
-<div class="issue" data-number="58" title="Token value data types">
-  JSON can support string, number, array, and object value types. Should all of these types be allowed for token values?
-</div>
+### Token value type
+
+Token values may be any valid JSON type:
+
+- `string`
+- `number`
+- `array`
+- `object`
+- `boolean`
+- `null`
+
+Additionally, tokens may be defined with a more specific [Token Type](#types)
 
 <div class="issue" data-number="55" title="Object vs Array">
 


c1rrus (approved):
LGTM 👍

technical-reports/format/design-token.md
```diff
@@ -23,9 +23,18 @@ The example above therefore defines 1 design token with the following properties
...
+### Token value type
+
+Token values may be any valid JSON type:
+
+- `string`
+- `number`
+- `array`
+- `object`
+- `boolean`
+- `null`
```
dbanksdesign, line 35: What would a `null` design token be?

c1rrus: I can't think if a good use for it. But, I also don't think that's a reason to artificially restrict what the underlying JSON format can do.

In practice, most (maybe all?) of the "low level", raw JSON types aren't especially useful to design token teams and tools. I mean, imagine throwing a value like `{ "foo": [1, true, 42, null], "bar": "baz" }` at Figma or StyleDictionary. It's valid JSON, but other than being able to say it's of type object, it's impossible to guess what it might represent (and I think it would be perfectly OK for tools to simply ignore tokens of types they can't do anything useful with).

I therefore expect the vast majority of tokens to be using the "higher level" DTCG types our spec defines (`color`, `dimension`, `duration`, etc.) which _do_ have clearly defined meanings.

However, still allowing the raw JSON types might serve as a sort of escape hatch for teams doing something exotic or rolling their own, bespoke tools. Especially, if they pair them up with `extensions`.

I'm keen to leave the door open and see what (if anything) folks do with them. It might give us some cowpaths to pave in future spec versions ;-)

dbanksdesign: That is a great explanation. I'm sold. Basically the design token spec supports all raw JSON types, even though they might not have a direct use in design tokens.

Issue #66, opened 2021-09-09
Consider using grammarly or hemingway to make language more understandable/inclusive


jina:
Comment: https://github.com/design-tokens/community-group/pull/52#pullrequestreview-748343934

kevinmpowell:
I'm going to start running the spec through hemingwayapp.com one section at a time.

kaelig:
Cross posting my comment in the PR above (https://github.com/design-tokens/community-group/pull/76#issuecomment-922128207) here – note that I'm a frequent user of both Hemingway and Grammarly, and encourage editors to use them.

---

I agree with the intent, but think an over-reliance on Hemingway is detrimental to the readability and clarity of this doc.

I find that following Hemingway's suggestions too closely leads to a succession of tiny sentences that can read as quite dry, and breaks the flow of otherwise very readable content. In one of the sentences, all jargon was removed in an undesirable lossy rewrite.

My guidance is to use Hemingway sparingly, and take its suggestions for what they are: suggestions.

👍🏻 It's useful to avoid very long sentences, repetitions, and spot passive voice that would read better as active.

👎🏻 It warns about jargon that's actually important in our context: technical reports are aimed at experts who rely on this vocabulary to derive precise meaning.

---

(Grammarly allows to specify the target audience, tone, level of formality… I personally find it better than Hemingway for professional writing like technical proposals, and still use Hemingway for blog posts)

jina:
That's fair. I wasn't necessarily saying we needed to use Hemingway. The purpose was to simplify the language a little bit, and I know those tools have helped me, so it was just a suggestion to consider a tool. I'm fine if we want to use just Grammarly!

christianmartinnies:
My five cents: We are trying to collaborate in between various disciplines here and from my personal experience i recommend, trying to be as precise as possible. I think, once we implemented a common language it will be beneficial—how to reach that is a different story on it's own.
So this is rather a matter of glossary but i would love to see maximum of inclusiveness as possible over here (heart)

PR #76, opened 2021-09-17
Fix/intro grammar

kevinmpowell:
I've used https://hemingwayapp.com to edit the intro section for conciseness. 

Here's the [before](https://deploy-preview-74--dtcg-tr.netlify.app/format/#status) and [after](https://deploy-preview-76--dtcg-tr.netlify.app/format/?utm_source=github&utm_campaign=bot_dp#status) for comparison.

References #66

### DIFF
--- a/technical-reports/format/index.html
+++ b/technical-reports/format/index.html
@@ -38,27 +38,29 @@
 
     <section id="abstract">
       <p>
-        This document describes the technical specification for a file format to
-        exchange design tokens between different tools.
+        This document describes the technical specification for a design tokens
+        file format. This format enables the exchange of design tokens between
+        different tools.
       </p>
     </section>
 
     <section id="status">
       <h1>Status of this document</h1>
       <p>
-        This is a snapshot of the editors’ draft. It is provided for discussion
-        only and may change at any moment. Its publication here does not imply
-        endorsement of its contents by W3C or the Design Tokens W3C Community
-        Group Membership. Don’t cite this document other than as work in
-        progress.
+        This is a snapshot of the editors’ draft. It is for discussion only and
+        may change at any moment. Its publication does not imply endorsement by
+        W3C or the Design Tokens W3C Community Group Membership. Don’t cite this
+        document other than as work in progress.
       </p>
-      <p>This document has been published to facilitate Wide Review.</p>
       <p>
-        This document was produced by the Design Tokens W3C Community Group, and
-        contributions to this draft are governed by
+        The Design Tokens W3C Community Group produced this document. They
+        published it to facilitate Wide Review.
+      </p>
+      <p>
+        The
         <a href="https://www.w3.org/community/about/process/cla">
           Community Contributor License Agreement (CLA) </a
-        >, as specified by the
+        >governs contributions to this draft, as specified by the
         <a href="https://www.w3.org/community/about/process/#cgroups"
           >W3C Community Group Process</a
         >.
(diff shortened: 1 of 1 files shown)


technical-reports/format/index.html
```diff
@@ -38,27 +38,29 @@
...
+      </p>
+      <p>
+        The
         <a href="https://www.w3.org/community/about/process/cla">
           Community Contributor License Agreement (CLA) </a
-        >, as specified by the
+        >governs contributions to this draft, as specified by the
         <a href="https://www.w3.org/community/about/process/#cgroups"
           >W3C Community Group Process</a
         >.
```
kaelig, line 66: This section is boilerplate from the W3C that you'll generally find untouched or very lightly edited in drafts. Not sure we should optimize for readability or consistency.

kaelig:
I agree with the intent, but think an over-reliance on Hemingway is detrimental to the readability and clarity of this doc.

I find that following Hemingway's suggestions too closely leads to a succession of tiny sentences that can read as quite dry, and breaks the flow of otherwise very readable content. In one of the sentences, all jargon was removed in an undesirable lossy rewrite.

My guidance is to use Hemingway sparingly, and take its suggestions for what they are: suggestions.

👍🏻 It's useful to avoid very long sentences, repetitions, and spot passive voice that would read better as active.

👎🏻 It warns about jargon that's actually important in our context: technical reports are aimed at experts who rely on this vocabulary to derive precise meaning.

---

(Grammarly allows to specify the target audience, tone, level of formality… I personally find it better than Hemingway for professional writing like technical proposals, and still use Hemingway for blog posts)

technical-reports/format/index.html
c1rrus: I'd lean towards consistency in this instance. After all, we're trying to align with W3C's tooling, process and writing style elsewhere.

technical-reports/format/index.html
```diff
@@ -74,57 +76,20 @@ <h1>Introduction</h1>
...
-        <li>
-          Documentation tools can display design token names rather than the raw
-          values in design specs and style guides.
-        </li>
-      </ul>
-      <p>
-        It is often desirable for design system teams to integrate such tools
-        together, so that design token data can flow between design and
-        development tools. For example:
+        Design tokens establish a common vocabulary across organizations. They
```
c1rrus, line 79: Whoa! Harsh edit :-P

I'm all for making things more concise and using plain English, but I think the specific examples of tools that might use tokens are useful here to set context.

technical-reports/format/index.html
```diff
@@ -74,57 +76,20 @@ <h1>Introduction</h1>
...
+        are a method for expressing design decisions.
       </p>
-      <ul>
```
c1rrus, line 106: If we leave the above bit as is, then we need to keep this list too.

technical-reports/format/index.html
```diff
@@ -74,57 +76,20 @@ <h1>Introduction</h1>
...
-        own, bespoke "glue" code or workflows. Furthermore, if teams want to
-        migrate to different tools, they will need to update those integrations.
+        Many tools allow importing and exporting design tokens, but formats are
+        tool-specific. The burden of sharing tokens across tools is on design
+        system teams. They must create and maintain their own, bespoke "glue"
+        code or workflows. Migrating to different tools means updating those
+        integrations.
       </p>
       <p>
-        This specification aims to facilitate better interoperability between
```
c1rrus, line 124: I think the original is better here. The edited version is too vague and "largest goal" just sounds weird to me.

c1rrus (changes requested):
I'd revert most of these edits TBH. They're stripping away too much info IMHO.

kevinmpowell:
Good feedback. Declining PR.

Issue #59, opened 2021-09-09
Token name case sensitivity


kevinmpowell:
## Should token names be case sensitive?
```
{
  colorTextPrimary: {
    value: '#000000'
  },
  ColorTextprimary: {
    value: '#000000'
  }
}
```

Should the above be two valid token keys (case-sensitive) or a duplicate entry (case-insensitive)?

mattfelten:
From a linting perspective, I'd want that example to give a warning. Thinking about a name translation standpoint, e.g camel case to kebab case, those two names would be translated differently, `color-text-primary` and `-color-textprimary` respectively, and I'd expect that. So they are unique to me.

My vote is case-sensitive

ariarzer:
I vote for the case-sensitive token's names. It will better for linting and consistency.
And, when it will translate for CSS custom property, for example, will be easy to linting a css-code too.

kevinmpowell:
> I'd want that example to give a warning

@mattfelten I completely agree. I think that may be sufficient. There may be environments that are case-insensitive where tokens would be used, but I can't think of any and those who operate in those environments can adapt as necessary.

dbanksdesign:
I agree with @kevinmpowell and @mattfelten case-sensitive, but the spec should include a warning about potential name collisions and how toolmakers should deal with it.

ChucKN0risK:
Reviewed by the spec editors on 2021-11-02

**Decision**
The spec should include a warning about potential name collisions and how toolmakers should deal with it.

**Action to be taken**
Next version of specification will reflect these decisions.

We’re closing this issue for now, but may reopen it in the future if needed.

Issue #55, opened 2021-09-08
Object vs Array


dbanksdesign:
```json
{
  "token name": {
    "value": "token value"
  }
}
```

The structure in the example above is a JSON object, an unordered set of name/value pairs.

- Objects can't contain members with duplicate keys
- Ordering of object members may not be preserved (as per [RFC 7159](https://tools.ietf.org/html/rfc7159#section-4)), meaning token retrieval may or may not result in the same ordering as the input

Please raise concerns if these limitations create problems for implementers.

c1rrus:
I believe it would useful if our format provided authors with a means to express that tokens in a group should be presented in a specific order. For example, if you have a bunch of tokens representing tints and shades of a colour then it's nice if they can be displayed in a specific order. Imagine a style guide generator that renders them as colour swatches.

Given that our basis is JSON, arrays feel like the most convenient way of achieving that. Doing that means that there are 2 flavours of groups:

* Object-based groups (whose tokens and nested groups do **not** have a defined order)
* Array-based groups (whose tokens and nested groups **do** have a defined order)

However, introducing array groups also brings a few challenges:

**Token names**
Right now, a token's name its key in the parent group. For example, to define a token called "my awesome token" you'd do something like this:
```jsonc
{
  "my awesome token": {
    "value": "#ff0000",
    "type": "color"
  }
}
```

But, if that's in an array, there is no (explicit) key:

```jsonc
[
  {
    // what is the name of this token??
    "value": "#ff0000",
    "type": "color"
  }
]
```

I had grappled with this a bit in my [UDT experiment](https://udt.design/), and landed on the following, which I propose as a potential solution for our spec:

* The token object itself can have an _optional_ `name` property to declare the token's name.
* If a token in an object has a `name` attribute, then that takes precedence over the parent object's key
* If a token in an array does _not_ have a `name` attribute, then the array index is used as its name

For example:

```jsonc
{
  "object group": {
    "token 1": {
      "value": 42
      // This token's name is: "token 1"
      // Its path is: "object group" / "token 1"
    }

    "token 2": {
       "name": "actual token name",
       "value": "hello world"
      // This token's name is: "actual token name" (note how the "token 2" key is now ignored)
      // Its path is: "object group" / "actual token name"
    }
  },

  "array group": [
    {
      "value": 99
      // This token's name is: "0" (since it has no name property, the array index is used instead)
      // Its path is: "object group" / "0"
    },

    {
      "value": 24,
      "name": "token name"
[...]

PR #83, opened 2021-11-15
Update spec to reflect decisions for #55 & #59

c1rrus:
* Mentions that items in a group are unordered (#55)
* Mentions that token names are case-sensitive (#59)
* Removes the RFC blocks for those issues

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -36,17 +36,6 @@ Token values may be any valid JSON type:
 
 Additionally, tokens may be defined with a more specific [Token Type](#types)
 
-<div class="issue" data-number="55" title="Object vs Array">
-
-The structure in the example above is a JSON object, an **unordered** set of name/value pairs.
-
-- Objects can't contain members with duplicate keys
-- Ordering of object members may not be preserved (as per [RFC 7159](https://tools.ietf.org/html/rfc7159#section-4)), meaning token retrieval may or may not result in the same ordering as the input
-
-Please raise concerns if these limitations create problems for implementers.
-
-</div>
-
 <div class="issue" data-number="59" title="Token name case sensitivity">
   Should token names be case sensitive?
 </div>
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -23,6 +23,35 @@ The example above therefore defines 1 design token with the following properties
 
 Name and value are both **required**.
 
+Token names are case-sensitive, so the following example with 2 tokens in the same group whose names only differ in case is valid:
+
+<aside class="example">
+
+```json
+{
+  "font-size": { "value": "3rem" },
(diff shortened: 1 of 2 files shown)


technical-reports/format/design-token.md
```diff
@@ -23,6 +23,35 @@ The example above therefore defines 1 design token with the following properties
...
+```json
+{
+  "font-size": { "value": "3rem" },
+  "FONT-SIZE": { "value": "16px" }
+}
+```
+
+</aside>
+
+However, some tools may need to transform names when exporting to other languages or displaying names to the user, so having token names that differ only in case is likely to cause identical and undesirable duplicates in the output. For example, an export tool that converts these tokens to SASS code may generate problematic output like this:
```
kaelig, line 39:
```suggestion
However, some tools may need to transform names when exporting to other languages or displaying names to the user, so having token names that differ only in case is likely to cause identical and undesirable duplicates in the output. For example, an export tool that converts these tokens to Sass code may generate problematic output like this:
```

c1rrus: Good spot. Sorry! After all these years of using Sass, I really ought to know better. Will push a fix in a moment.

technical-reports/format/design-token.md
```diff
@@ -23,6 +23,35 @@ The example above therefore defines 1 design token with the following properties
...
+$font-size: 3rem;
+$font-size: 16px;
+
+// The 2nd $font-size overrides the 1st one, so the 1st token
+// has essentially been lost.
+```
+
+</aside>
+
+Tools therefore MAY display a warning to the user when encountering token names that may cause naming clashes after being exported.
```
kevinmpowell, line 53: Alternative:

Tools MAY display a warning when token names differ only by case.

technical-reports/format/groups.md
```diff
@@ -48,6 +48,10 @@ The names of the groups leading to a given token (including that token’s name)
 
 Because groupings are arbitrary, tools MUST NOT use them to infer the type or purpose of design tokens.
 
+Groups are unordered. In other words, there is no implicit order between items (i.e. the tokens and/or nested groups) within a group. Therefore, tools that parse or write design token files are not required to preserve the source order of items in a group.
```
kevinmpowell, line 51: Suggested edit: 

> Group **items** are unordered. In other words, there...

kevinmpowell (approved):
Looks good @c1rrus, just a couple small suggestions for clarity.

ChucKN0risK (approved):

dbanksdesign (approved):
LGTM! :shipit:

c1rrus:
Thanks folk!  I'll push a small commit with @kevinmpowell's suggested edits later tonight.

Do y'all want to check that change once I've done it, or are you happy for me to merge the PR then?

kevinmpowell:
Merge when you're ready, I don't need to see it again.

Issue #63, opened 2021-09-09
Token types - optional or required


kevinmpowell:
## Are token types optional or required?

Example:
```
{
  colorTextPrimary: {
    value: '#000000',
    type: 'Color'
  }
}
```

nhoizey:
I would vote for optional, as I currently prefer using one file per token type, such as `color.json`, so repeating the same type for every token would not be ideal.

mattfelten:
I don't like requiring maintainers to do more work to keep their tokens working so my initial thought was optional.

Requiring it now will allow consuming tools more consistent and easier time implementing tokens, which to me is a big win. Not having types would mean each tool will have to parse and type-check themselves which could lead to inconsistent results tool to tool.

My vote is required.

blackfalcon:
> Not having types would mean each tool will have to parse and type-check themselves

I disagree. Why do I have to meet a specification for a tool I may not be using? If I am using it, there would be a requirement from that tool to have the `type` defined. That's my only motivation to do so, otherwise, I fall in the same camp as @nhoizey.

jina:
note, there are tools like Theo that let you set a type globally for all the tokens in a file, so it doesn't have to be on each individual token. if we allowed that to fulfill the requirement, would that help? I think types are pretty important.

blackfalcon:
The question is, is the enforcement of the type for the betterment of the library or for the ease of integration of a 3rd party tool? Or maybe better said, is this still useable without the mandatory type? While I agree typing can be helpful, what is the driver for this? 

Unless there is a direct benefit to the lib itself, making this required is inappropriate IMHO.

jina:
@blackfalcon I'm not 100% sure what you mean by "the lib" (the token output? the framework you're importing them in?)

As aligned with our mission objectives, we want to be **extensible**. Design tokens are also meant to be agnostic so that they can be usable by various libs and tools. Part of our goal is to give design tools/vendors a way to integrate tokens in a predictable way that makes sense and has agreement with the community. This was a big motivator in why we established this community group in the first place. Types seem necessary to make this work. 

And in terms of the "lib", if you mean your design system codebase, how are you documenting that? I find that types being required help display tokens in the appropriate way in the docs (colors vs units vs whatever else). Additionally, when using JS/TypeScript or other languages beyond CSS/Sass, I've found it can be very useful for linting/testing/validation (which hopefully you're doing!) and testing certain attributes for accessibility.

[...]

PR #86, opened 2021-12-05
Update composite types chapter to reflect decision made on issue #63

### LINKED ISSUE #54: Should composites be part of the MVP specification?
If so, which composites should be included initially?
- border
- transition
- shadow
- gradient
- text style
- color pair

Other questions
- Would composites allow for better integration with design tools?
- How would user-defined composites be rendered within design tools?

c1rrus:
Closes #54 

The format editors have decided to keep the concept of composite types in the draft spec. However, we will replace the mechanism for user-defined composite types with a set of pre-defined composite types. (Future spec versions may bring back user-defined (composite) types if there is demand for it).

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -95,7 +95,34 @@ Represents a border style. The type property must be set to the string “border
 
 - `color`: The color of the border. The value of this property must be a valid [color value](#color) or a reference to a color token.
 - `width`: The width or thickness of the border. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
-- `style`: The border's style, for example "solid" or "dashed". The value of this property must be a valid JSON string or a reference to another string token.
+- `style`: The border's style, for example "solid" or "dashed". The value of this property must be a valid JSON string or a reference to a string token.
+
+<aside class="example" title="Border composite token examples">
+
+```json
+{
+  "border": {
+    "heavy": {
+      "type": "border",
+      "value": {
+        "color": "#36363600",
+        "width": "3px",
+        "style": "solid"
+      }
+    },
+    "focusring": {
+      "type": "border",
+      "value": {
+        "color": "{color.focusring}",
+        "width": "1px",
+        "style": "dashed"
+      }
+    }
+  }
+}
+```
+
+</aside>
 
 # Transition
 
@@ -115,6 +142,43 @@ Represents a shadow style. The type property must be set to the string “shadow
 
 TO-DO
 
-# Text style
+# Typography
 
-TO-DO
+Represents a typographic style. The type property must be set to the string “typography”. The value must be an object with the following properties:
+
+- `fontName`: The typography's font. The value of this property must be a valid [font name](#font-name) or a reference to a font name token.
+- `fontSize`: The size of the typography. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `fontWeight`: The weight of the typography. The value of this property must be a valid JSON string or a reference to a string token.
+- `letterSpacing`: The horizontal spacing between characters. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `lineHeight`: The vertical spacing between lines of typography. The value of this property must be a valid JSON string or a reference to a string token.
+
+<aside class="example" title="Typography composite token examples">
+
+```json
+{
+  "type styles": {
+    "heading-level-1": {
+      "type": "typography",
+      "value": {
+        "fontName": "Roboto",
+        "fontSize": "42px",
+        "fontWeight": "700",
+        "letterSpacing": "0.1px",
+        "lineHeight": "1.2"
+      }
+    },
+    "microcopy": {
+      "type": "typography",
+      "value": {
+        "fontName": "{font.serif}",
+        "fontSize": "{font.size.smallest}",
+        "fontWeight": "{font.weight.normal}",
+        "letterSpacing": "0px",
+        "lineHeight": "1"
+      }
+    }
+  }
+}
+```
+
+</aside>
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -6,13 +6,14 @@ Every shadow style has the exact same parts (color, X & Y offsets, etc.), but th
 
 Specifically, a composite type has the following characteristics:
 
-- Its value is an object containing sub-values.
-- Each sub-value has a pre-defined name and type.
+- Its value is an object or array containing sub-values.
+- For object values, each sub-value has a pre-defined name and type.
+- For array values, all elements of the array are sub-values that have the same pre-defined type.
 - Sub-values may be explicit values (e.g. `"#ff0000"`) or references to other design tokens that have sub-value's type (e.g. `"{some.other.token}"`).
 
 A design token whose type happens to be a composite type is sometimes also called a composite (design) token. Besides their type, there is nothing special about composite tokens. They can have all the other additional properties like [description](#description) or [extensions](#extensions). They can also be referenced by other design tokens.
 
-<aside class="example" title="Composite token exmple">
+<aside class="example" title="Composite token example">
 
 ```json
 {
@@ -138,9 +139,107 @@ Represents a shadow style. The type property must be set to the string “shadow
 - `blur`: The blur radius that is applied to the shadow. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
 - `spread`: The amount by which to expand or contract the shadow. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
 
+# Gradient Stop
+
+Represents an individual stop on a color gradient. In practice, this type is unlikely to be useful by itself, but it is required by the [gradient type](#gradient). The value must be an object with the following properties:
+
+- `color`: The color value at the stop's position on the gradient. The value of this property must be a valid [color value](#color) or a reference to a color token.
+- `position`: The position of the stop along the gradient's axis. The value of this property must be a valid number value or reference to a number token. The number values must be in the range [0, 1], where 0 represents the start position of the gradient's axis and 1 the end position. If a number value outside of that range is given, it MUST be considered as if it were clamped to the range [0, 1]. For example, a value of 42 should be treated as if it were 1, i.e. the end position of the gradient axis. Similarly, a value of -99 should be treated as if it were 0, i.e. the start position of the gradient axis.
+
 # Gradient
 
-TO-DO
+Represents a color gradient. The value must be an array of [gradient stops](#gradient-stop). If there are no stops at the very beginning or end of the gradient axis (i.e. with `position` 0 or 1, respectively), then the color from the stop closest to each end should be extended to that end of the axis.
+
+<aside class="example" title="Gradient token example">
+
+```json
+{
+  "blue-to-red": {
+    "type": "gradient",
+    "value": [
+      {
+        "color": "#0000ff",
+        "pos": 0
+      },
+      {
+        "color": "#ff0000",
+        "pos": 1
+      }
+    ]
+  }
+}
+```
+
+Describes a gradient that goes from blue to red:
+
+<div style="height: 2rem; background: linear-gradient(90deg, #0000ff, #ff0000);"></div>
+
+</aside>
+
+<aside class="example" title="Gradient token with omitted start stop example">
+
+```json
+{
+  "mostly-yellow": {
+    "type": "gradient",
+    "value": [
+      {
+        "color": "#ffff00",
+        "pos": 0.666
+      },
+      {
+        "color": "#ff0000",
+        "pos": 1
+      }
+    ]
+  }
+}
+```
+
+Describes a gradient that is solid yellow for the first 2/3 and then fades to red:
+
+<div style="height: 2rem; background: linear-gradient(90deg, #ffff00 66.6%, #ff0000);"></div>
+
+</aside>
+
+<aside class="example" title="Gradient token using references example">
+
+```json
+{
+  "brand-primary": {
+    "type": "color",
+    "value": "#99ff66"
+  },
+
+  "position-end": {
+    "value": 1
+  },
+
+  "brand-in-the-middle": {
+    "type": "gradient",
+    "value": [
+      {
+        "color": "#000000",
+        "pos": 0
+      },
+      {
+        "color": "{brand-primary}",
+        "pos": 0.5
+      },
+      {
+        "color": "#000000",
+        "pos": "{position-end}"
+      }
+    ]
+  }
+}
+```
+
+Describes a color token called "brand-primary", which is referenced as the mid-point of a gradient is black at either end.
+
+<div style="height: 2rem; background: linear-gradient(90deg, #000000, #99ff66, #000000);"></div>
+
+</aside>
 
 # Typography
 
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -19,8 +19,8 @@ A design token whose type happens to be a composite type is sometimes also calle
     "type": "shadow",
     "value": {
       "color": "#00000088",
-      "x": "0.5rem",
-      "y": "0.5rem",
+      "offset-x": "0.5rem",
+      "offset-y": "0.5rem",
       "blur": "1.5rem",
       "spread": "0rem"
     }
@@ -54,8 +54,8 @@ A design token whose type happens to be a composite type is sometimes also calle
       "description": "A composite token where some sub-values are references to tokens that have the correct type and others are explicit values",
       "value": {
         "color": "{color.shadow-050}",
-        "x": "{space.small}",
-        "y": "{space.small}",
+        "offset-x": "{space.small}",
+        "offset-y": "{space.small}",
         "blur": "1.5rem",
         "spread": "0rem"
       }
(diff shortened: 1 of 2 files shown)


kevinmpowell:
@c1rrus I'm not convinced we need the `GradientStop` type. 
I suggest Gradient's type definition to be "an array of objects with `Color` and `Number` values"

What value does the `GradientStop` provide?

c1rrus:
@kevinmpowell Yeah, I wasn't too keen on adding the `gradient-stop` type myself. My thinking was as follows: In the composite type intro, we say that composite type values are either objects whose properties are sub-values, or they are an array of sub-values. Sub-values in turn are defined as a value from one of the types or a reference to a token of the same type.

Therefore, having an array of objects which themselves are _not_  sub-values, but whose properties _are_ doesn't quite fit that definition. Breaking it up into 2 types gets around that.

But... it's klunky.

So, I'll have a go at rewording our definition of composite types so that it's broader. Then I can get rid of `gradient-stop` as its own type.

c1rrus:
@kevinmpowell Ok, just pushed an update that merges the gradient stop stuff into the gradient type.

technical-reports/format/composite-types.md
```diff
@@ -60,155 +75,205 @@ Other examples might be:
...
+Represents a typographic style. The type property must be set to the string “typography”. The value must be an object with the following properties:
 
-Via configuration files or settings menus, tools could let users define custom behavior for types they have defined. For instance, in an export tool like Style Dictionary, users might be able to configure a custom transform or formatter for their composite types.
-Then, this example...
+- `fontName`: The typography's font. The value of this property must be a valid [font name](#font-name) or a reference to a font name token.
+- `fontSize`: The size of the typography. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `fontWeight`: The weight of the typography. The value of this property must be a valid JSON string or a reference to a string token.
+- `letterSpacing`: The horizontal spacing between characters. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `lineHeight`: The vertical spacing between lines of typography. The value of this property must be a valid JSON string or a reference to a string token.
```
c1rrus, line 248: Wouldn't a number or dimension (or both) be more appropriate here?

Otherwise, if it's a string value, then what's the expected syntax within that string value?

technical-reports/format/composite-types.md
```diff
@@ -60,155 +75,205 @@ Other examples might be:
...
+## Typography
 
-### Custom configuration
+Represents a typographic style. The type property must be set to the string “typography”. The value must be an object with the following properties:
 
-Via configuration files or settings menus, tools could let users define custom behavior for types they have defined. For instance, in an export tool like Style Dictionary, users might be able to configure a custom transform or formatter for their composite types.
-Then, this example...
+- `fontName`: The typography's font. The value of this property must be a valid [font name](#font-name) or a reference to a font name token.
+- `fontSize`: The size of the typography. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `fontWeight`: The weight of the typography. The value of this property must be a valid JSON string or a reference to a string token.
```
c1rrus, line 246: Should we restrict this to a pre-defined set of values (effectively, a sort of enum). E.g. `"bold"`, `"normal"`, `"light"`, etc. and/or numeric values like `100`, `200`, `300`, etc.?

Otherwise, any string is technically valid, so how should a tool interpret something like `"fontWeight": "hello"`?

If we do limit the values, I think it might make sense to define that as its own type (maybe "font-weight" or something like that)? It's conceivably something people might also define individual tokens for.

technical-reports/format/composite-types.md
```diff
@@ -60,155 +75,205 @@ Other examples might be:
...
+
+## Border
 
-## Type checking
+Represents a border style. The type property must be set to the string “border”. The value must be an object with the following properties:
 
-Just as with “normal” types for tokens, using a custom, composite type will allow tools to check that the values you use match the expected type. In our color pair example above, attempting to do the following would be invalid and tools should ignore the token and show an error to the user, since “Comic Sans MS” is not a valid color value.
+- `color`: The color of the border. The value of this property must be a valid [color value](#color) or a reference to a color token.
+- `width`: The width or thickness of the border. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `style`: The border's style, for example "solid" or "dashed". The value of this property must be a valid JSON string or a reference to a string token.
```
c1rrus, line 97: Similar to font weight on the typography type, I wonder whether there should be a stand-alone `border-style` type that defines a set of known, permitted values (e.g. `"solid"`, `"dashed"`, `"dotted"`, etc.).

Otherwise any string would be valid and its unclear how tools should interpret the value

technical-reports/format/composite-types.md
ChucKN0risK: @drwpow Thanks a lot for sharing your thoughts on this. The more feedback we get from the community the better it is tbh.

I agree with @drwpow regarding the `fontWeight` property. It's a tricky one because there's no clear convention.

In my opinion its value should be restricted to a set of string and numeric values.

String values:
- `ultrathin`
- `thin`
- `ultralight`
- `light`
- `regular`
- `book`
- `normal`
- `bold`
- `ultrabold`

I might have forgotten other frequently used values here though. The tricky part is that there's no strict convention between typefaces for these values.

Numeric values from `1` to `999`.

technical-reports/format/composite-types.md
ChucKN0risK: I agree. Its value should be restricted to a set of the following strings: `solid`, `dashed`, `dotted`.

technical-reports/format/composite-types.md
c1rrus: Inspired by @ChucKN0risK & @drwpow's comments above, I propose the following:

* We add a new (non-composite) `fontWeight` type
* The `fontWeight` sub-value of the `typography` composite type must have values that are valid `fontWeight` token values or references to `fontWeight` tokens.

The new `fontWeight` type's values must be either a number value in the range `1` - `1000` (which seems to be the range [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide#weight) and [Android](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/font/FontWeight) permit) _or_ one of the pre-defined strings listed in the table below. As per the table, the string names are just human-friendly aliases for certain number values, just like [how CSS does it](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping), which in turn seems to be based on the [OpenType spec](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#wtc):

| numeric value | string value aliases |
|-----------------|--------------------- |
| `100` | `thin`, `hairline` |
| `200` | `extraLight`, `ultraLight` |
| `300` | `light` |
| `400` | `normal`, `regular`, `book` | 
| `500` | `medium` |
| `600` | `semiBold`, `demiBold` |
| `700` | `bold` |
| `800` | `extraBold`, `ultraBold` |
| `900` | `black`, `heavy` |
| `950` | `extraBlack`, `ultraBlack` |

Sound good?

technical-reports/format/composite-types.md
c1rrus: Hmm... on second thoughts, a `borderStyle` (or perhaps `strokeStyle` is a better name?) type might need to be more complex. I've done a little bit of research and there's a few ways different platforms and tools do this:

* CSS has a finite number of ["line style" keywords](https://developer.mozilla.org/en-US/docs/Web/CSS/border-style#line-style) (`solid`, `dashed`, `dotted`, etc.), but doesn't offer finer control beyond that. 
* SVG has an attribute called [`stoke-dasharray`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray), which lets you define a patterns of dash and gap lengths. There's also a [`stroke-linecap`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap) for round / butt / square caps.
* Figma has "Solid" / "Dash" / "Custom" styles. For dash and custom, you can set the dash & gap lengths in pretty much the same way as SVG. It also lets you choose from the same 3 linecap styles that SVG has, except they call it "Dash cap".
* Android seems to let you set [dash width and dash gap](https://developer.android.com/guide/topics/resources/drawable-resource#stroke-element) (so equivalent to Figma's "Dash" style), but doesn't support custom patterns like SVG & Figma do

This is of course not an exhaustive list (if anyone knows what iOS, Windows or other things support, then please share), but I think it's safe to say that a stroke style type is non-trivial.

Perhaps `strokeStyle` is itself a composite type? For example, we might define its values as objects with the following properties:

* `dashArray`: An array of `dimension` values, which could also be references to dimension tokens (as SVG does it)
* `lineCap`: One of the following strings `"round"`, `"butt"`, `"square"`

For example:

```json
{
  "line-dashed": {
    "type": "strokeStyle",
    "value": {
      "dashArray": ["3px", "{size-medium}"],
      "lineCap": "round"
    }
  },
  
  "size-medium": {
    "type": "dimension",
    "value": "5px"
  },
  
  "button-border": {
    "type": "border",
    "value": {
      "color": "#ff0000",
      "width": "{size-medium}",
      "style": "{line-dashed}"
    }
  }
}
```

But, similar to what I suggested for `fontWeight` in an earlier PR comment, we may _also_ want to have a set of keywords (similar to CSS) for common stroke styles. In that case, we ought to define some kind of mapping between the two. For example we might define `"dashed"` to be an alias for `{ "dashArray": ["3px", "3px"], "lineCap": "butt" }`. In that case, the following would also be valid:

```json
{
  "line-dashed": {
    "type": "strokeStyle",
    "value": "dashed"
  }
}  
```

One catch here is that the CSS line styles seem to scale relative to the border's width. So, the dashes and gaps on a `10px` border are wider than they are on a `1px` border. As far as I can tell, in SVG and Figma this is _not_ the case - the dash and gap values are essentially absolute lengths that remain the same regardless of how wide the border is. I threw together a little [CodePen to illustrate this](https://codepen.io/cirrus/pen/poWWBEO). At least in SVG, I couldn't figure out a way to make a dasharray scale relative to the stroke width. In other words SVG/Figma can do stuff CSS can't and CSS and do stuff that SVG/Figma can't! I have no idea what Android or iOS do in this respect (maybe @dbanksdesign knows?).

So maybe a mapping table isn't feasible?

-----

On a separate note, going down this route would also be setting an interesting precedent, since we'd have a token type that is both a composite and non-composite type in one! I can think of other cases where that might be useful. Do we need a name for that? "Hybrid type" perhaps? ;-)

technical-reports/format/composite-types.md
kevinmpowell: CSS often uses unitless line heights, which would push me toward "number" for the type, but that would exclude the definition of a line height that includes the unit, hence "string"

Our current dimension Type requires either `px` or `rem` excluding the possibility of using unitless line heights,

technical-reports/format/composite-types.md
kevinmpowell: I support this proposal @c1rrus

technical-reports/format/composite-types.md
kevinmpowell: @c1rrus thanks for all the research you did on this. I think a `strokeStyle` composite type is needed, but should have the shorthand you're describing to align with CSS behaviors. 

I don't think we need a separate name for this type of token. It's still a composite token, it just has a shorthand for setting some of its sub values.

technical-reports/format/composite-types.md
c1rrus: Great! I've just pushed a commit that adds it and updates the `typography` type accordingly.

I went for a kebab case `font-weight` name rather than camel case as I used in my comment above in order to be consistent with the `cubic-bezier` type we already have. I also renamed the existing `font` type to `font-name` to make its purpose less ambiguous.

I wonder whether we should also change the sub-value names we've used in composite types to be kebab-case for the sake of consistency (or else, make everything camel case). What do you think?

technical-reports/format/composite-types.md
c1rrus: Why not both? Could we not say the `lineHeight` sub-value must be a valid dimension _or_ number value?

Perhaps the ability for sub-types to be one of several types could be (yet) another special quality of composite types?

technical-reports/format/composite-types.md
c1rrus: +@kevinmpowell

c1rrus:
+@kevinmpowell

technical-reports/format/composite-types.md
```diff
@@ -60,155 +75,205 @@ Other examples might be:
...
-## Tool support
+## Shadow
 
-Since composite types are user-defined, tools cannot provide specialized handling of them since they have no advanced knowledge of the type declarations. Here are some suggested strategies to get the most value out of user-defined composite types:
+Represents a shadow style. The type property must be set to the string “shadow”. The value must be an object with the following properties:
 
-### Fallbacks
+- `color`: The color of the shadow. The value of this property must be a valid [color value](#color) or a reference to a color token.
+- `x`: The horizontal offset that shadow has from the element it is applied to. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
+- `y`: The vertical offset that shadow has from the element it is applied to. The value of this property must be a valid [dimension value](#dimension) or a reference to a dimension token.
```
dbanksdesign, line 136: Although this might be more verbose, should we call `x` and `y` `offsetX` and `offsetY` respectively? The CSS spec and iOS use the term `offset`:
* https://drafts.csswg.org/css-backgrounds/#vertical-offset
* https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#values
* https://developer.apple.com/documentation/quartzcore/calayer/1410970-shadowoffset

technical-reports/format/composite-types.md
dbanksdesign: `lineHeight` could be a type itself, which is a union type of: number, dimension, percentage, or keyword 'normal' (based on the CSS spec at least: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height)

technical-reports/format/composite-types.md
```diff
@@ -1,57 +1,72 @@
 # Composite types
 
-<div class="issue" data-number="54" title="Should composites be part of the MVP specification?">
+The types defined in the previous chapters such as color and dimension all have singular values. For example, the value of a color token is _one_ color. However, there are other aspects of UI designs that are a combination of multiple values. For instance, a shadow style is a combination of a color, X & Y offsets, a blur radius and a spread radius.
 
-If so, which composites should be included initially?
+Every shadow style has the exact same parts (color, X & Y offsets, etc.), but their respective values will differ. Furthermore, each part's value (which is also known as a "sub-value") is always of the same type. A shadow's color must always be a [color](#color) value, its X offset must always be a [dimension](#dimension) value, and so on. Shadow styles are therefore combinations of values _that follow a pre-defined structure_. In other words, shadow styles are themselves a type. We call types like this **composite types**.
```
dbanksdesign, line 5: Minor: should we not use "we" in the spec? "We call types like this..." v. "Types like this are called..."

dbanksdesign (commented):
Looking good! Added a few comments

technical-reports/format/composite-types.md
ChucKN0risK: I think we should allow users to define unitless line heights. Hence, I agree with @dbanksdesign and I think we could add a `lineHeight` type.

technical-reports/format/composite-types.md
ChucKN0risK: > I wonder whether we should also change the sub-value names we've used in composite types to be kebab-case for the sake of consistency (or else, make everything camel case). What do you think?
@c1rrus You're suggesting to rename `extraLight` as `extra-light`?
If so, I agree.

technical-reports/format/composite-types.md
ChucKN0risK: I suggest we shouldn't use "we" in the spec.

c1rrus: Agreed. I'll fix that.

technical-reports/format/composite-types.md
c1rrus: @ChucKN0risK not quite. I already used kebab case for the values (e.g. `extra-light`). But the names of the typography type's sub-values are currently camel case:

```jsonc
{
  "example token": {
    "type": "typography",
    "value": {
      "fontWeight": "extra-light" // <-- this looks weird to me

      // ...
    }
  }
}
```
So, what I am proposing is that we make those kebab-case too: `"font-weight": "extra-light"`.

Furthermore, we should then adopt this convention across our entire format.

technical-reports/format/composite-types.md
c1rrus: @kevinmpowell I've just pushed a commit that adds a new `stroke-style` composite type and updates the `border` type to use that.

technical-reports/format/composite-types.md
c1rrus: Done

c1rrus:
At the last format editors' call we decided to merge this PR more or less as-is and then open issues for each of the new composite types, so that we could have more focussed discussions for each and, where needed, open separate PRs to refine them.

I have therefore opened a bunch of new GitHub issues to collect feedback and pushed a commit that:

- adds an initial stab at the missing transition type (so that there's something for folks to critique rather than it being completely open)
- adds links in each composite type section to the respective GitHub issue

I suggest we do a final review of this PR (focussing more on things like grammar, typos, etc. rather than the content itself) so that we can get this merged. Discussion and work on the various composite types can then continue in the new issues.

technical-reports/format/composite-types.md
natemoo-re: +1 to `kebab-case` everywhere

technical-reports/format/composite-types.md
```diff
@@ -169,46 +235,221 @@ For example, imagine Sketch wanted to export all layer styles from a Sketch docu
...
+
+<div class="issue" data-number="101" title="Gradient type feedback">
+  Is the current specification for gradients fit for purpose? Does it need to also specify the type of gradient (.e.g linear, radial, concial, etc.)?
+</div>
+
+## Typography
+
+Represents a typographic style. The type property must be set to the string “typography”. The value must be an object with the following properties:
+
+- `fontName`: The typography's font. The value of this property must be a valid [font name](#font-name) or a reference to a font name token.
```
drwpow, line 414: In the interest of sticking to `kebab-case`, what about naming this `font-family`? Otherwise it’d be the only property in `typography` that doesn’t map 1:1 to CSS.

Also, being pedantic, it technically is referring to a family because the name would include weight + italic/regular

c1rrus: Good point, thanks! @mkeftz made [the same observation](https://github.com/design-tokens/community-group/issues/102#issuecomment-1012610832) in #102.

I'll change it to `fontFamily` (at the last editor's call we decided to standardise on camel case, so I might as well change. it accordingly now).

dbanksdesign (approved):
LGTM! This is probably one of the biggest PRs we've had to date and represents some amazing work. Thank you @c1rrus for driving this!

thank you

PR #87, opened 2021-12-13
update(spec): Add `type` property to file and group levels

ChucKN0risK:
Update spec to reflect decision: https://github.com/design-tokens/community-group/issues/63#issuecomment-946936561

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -14,12 +14,12 @@
 
 </aside>
 
-An object with a “**value**” property is a token. Thus, “value” is a reserved word in our spec, meaning you can’t have a token whose name is “value”. The parent object’s key is the token name.
+An object with a "**value**" property is a token. Thus, "value" is a reserved word in our spec, meaning you can't have a token whose name is "value". The parent object's key is the token name.
 
 The example above therefore defines 1 design token with the following properties:
 
-- Name: “token name”
-- Value: “token value”
+- Name: "token name"
+- Value: "token value"
 
 Name and value are both **required**.
 
@@ -75,11 +75,13 @@ Additionally, tokens may be defined with a more specific [Token Type](#types)
 
 ## Additional properties
 
-While “value” is the only required property for a token, a number of additional properties may be added:
+While "value" is the only required property for a token, a number of additional properties may be added:
 
 ## Description
 
-A plain text description explaining the token’s purpose. Tools MAY use the description in various ways. For example:
+A plain text description explaining the token's purpose. Tools MAY use the description in various ways.
+
+For example:
 
 - Style guide generators could display the description text alongside a visual preview of the token
 - IDEs might display the description as a tooltip for auto-completion (similar to how API docs are displayed)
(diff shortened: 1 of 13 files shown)


technical-reports/format/design-token.md
```diff
@@ -107,17 +109,35 @@ The **description** property must be a plain JSON string, for example:
 
 ## Type
 
-Declares the type of the token. [See “Types”](#types) for more information.
+Declares the type of the token. [See supported "Types"](#types) for more information.
```
c1rrus, line 113: Maybe we should add a bit more info to make it clear that while the `type` property is optional, every token does have an unambiguous type.

Perhaps something like this:

```suggestion

Design tokens always have an unambiguous type, so that tools can reliably interpret their value. If a token's type is not explicitly specified via the `type` property, then the token's type MUST be determined as follows:

* If the token's value is a reference, then its type is the type of the token being referenced.
* Otherwise, if any of the token's parent groups have a `type` property, then the token's type is inherited from the closest parent group with a `type` property.
* Otherwise, the token's type whichever of the basic JSON types (`string`, `number`, `boolean`, `object`, `array` or `null`) its value is.

```

technical-reports/format/groups.md
```diff
@@ -95,11 +97,37 @@ Groups may support additional properties like type and description. Should other
 
 </div>
 
+### Type
+Groups may include an optional `type` property so a type property does not need to be manually added to every token. [See supported "Types"](#types) for more information.
+
```
c1rrus, line 102: I think we should add something here to explain how the `type` property on a group should be interpreted. For example:

```suggestion

If a group has a `type` property it acts as a default type for any tokens within the group, including ones in nested groups, that do not explicity declare a type via their own `type` property. For the full set of rules by which a design token's type is determined, please refer to the [design token type property chapter](#type-0).

```

c1rrus (changes requested):
Looking good. I've suggested a few additions to help clarify how the `type` properties should be interpreted.

(Also thanks for cleaning up the quotes. I know I agreed to doing this now but, in hindsight, it might have been better to do that in a separate PR so that the actual content changes are easier to find in the diff. Maybe that's how we should do it in future :-) )

technical-reports/format/design-token.md
```diff
@@ -107,17 +109,35 @@ The **description** property must be a plain JSON string, for example:
...
 
-<div class="issue" data-number="63" title="Token types optional or required">
-  Are token types optional or required?
-</div>
+The `type` property can be set on different levels:
+- at the file level
+- at the group level
+- at the token level
+
+The `type` property must be a plain JSON string.
```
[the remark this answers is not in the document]
kevinmpowell: @ChucKN0risK we discussed this in our format group meeting and decided we should remove the `- at the file level` bullet, since we haven't discussed how multiple files would work.

The root of the file is explicitly a group.

technical-reports/format/groups.md
ChucKN0risK: I've updated the PR to include your suggestion @c1rrus.

technical-reports/format/design-token.md
ChucKN0risK: I've updated the PR to include your suggestion @c1rrus.

technical-reports/format/design-token.md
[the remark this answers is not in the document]
ChucKN0risK: Thanks @c1rrus and @kevinmpowell!

I've updated the PR to:
- include your suggestion @c1rrus 
- remove `- at the file level bullet`

c1rrus (approved):
Aside from the Prettier linting which seems to be failing on the CI build, I'm happy to approve this.

Would be good to get this merged soon since it touches so many files. That'll make it easier for folks to do follow-up PRs with less chance of merge conflicts later on.

kevinmpowell:
I just pushed a commit that should resolve the prettier errors. I agree with @c1rrus, we should merge this ASAP. @ChucKN0risK can you merge when everything passes?

ChucKN0risK:
Done @kevinmpowell

Issue #72, opened 2021-09-14
Group & file level properties


kevinmpowell:
Several issues have raised the need for adding token properties at group and file levels. I'm creating this issue to consolidate discussions about this topic.

For example Types:

```js
{
  "colorTextPrimary": {
    "value": "#000000",
    "type": "Color"  // Tedious
  },
  "colorTextSecondary": {
    "value": "#333333",
    "type": "Color" // Tedious
  },
  ...
}


///  VS  ///
{
  "color": {
    "type": "Color", // Define once for the grouping
    "textPrimary": {
      "value": "#000000",
    },
    "textSecondary": {
      "value": "#333333",
    },
  ...
  }
}
```

Or descriptions for documentation about an entire group:

```js
{
  "colorText": {
    "type": "Color",
    "description": "Our palette of colors for text only.",
    "primary": {
      "value": "#000000",
      "description": "Use as the default text color",
    },
    "secondary": {
      "value": "#333333",
      "description": "Use for text that is not as important."
    },
  ...
  }
}
```

c1rrus:
I'm definitely in favour of allowing both `description` and `type` on groups as well as tokens. Furthermore, if/when we add any other properties in future we should always consider if they are applicable to tokens, groups or both.


Note, that with `type`, I think its meaning when applied to a group should be subtly different to when it's applied to a token:

* `type` on a token is explicitly stating what type that token is.
* `type` on a group is setting a default type for all tokens within that group (and any nested groups) which applies unless the tokens have their own `type` property, which then takes precedence.

Example:

```jsonc
{
  "group 1": {
    "type": "color",

    "token A": {
      // This token has the type color because it inherits it from group 1
      "value": "#00ff00"
    },

    "token B": {
      // This token has the type dimension, because its own type property
      // takes precedence the inherited color type
      "type": "dimension",
      "value": "12rem"
    },

    "group 1.1": {
      "token B": {
        // This token also has type color because it inherits it from group 1
        // since that is the nearest group with a type property
        "value": "#ff0000"
      }
    },

    "group 1.2": {
      "type": "duration",
      
      "token C": {
        // This token has type duration because it inherits it from group 1.2
        // (group 1.2's type takes precedence over its parent group's type)
        "value": "150ms"
      }
  },

  "token X": {
    // This token is not within a group with a type, so it does not inherit a type.
    // It also does not have its own type property. Therefore, tools must fall back
    // to the JSON type of its value, which is "string".
[...]

Issue #61, opened 2021-09-09
Token name - reserved words


kevinmpowell:
## Are there any reserved words that should not be used in token names?

Examples: 'default', 'int'
```
{
  default: {
    value: '#000000'
  },
  int: {
    value: 50
  }
}
```

Will these (and others) cause problems when converted to various languages and platforms?

jina:
Hmm. We use `default` quite a bit at work! 😛

blackfalcon:
The purpose of reserved words is to keep users from colliding with words that have a specific meaning to the language. Is there a language taxonomy for this project? Like @jina pointed out, people use `default`. 

I am curious why you picked  `default` as an example. Or `int`, what would that collide with here?

kevinmpowell:
I've seen `default` create collisions in the Swift language on iOS.
See: https://stackoverflow.com/questions/49352629/why-cant-i-put-an-enum-case-in-single-quotes-which-is-of-type-int

jina:
oh interesting. for some reason that's not been an issue for our ios tokens yet. I'll bring it up with our team just in case. :)

c1rrus:
There's a bunch of property names that have a special meaning in our spec, so I believe these necessarily need to be reserved words: `value`, `type`, `description`, `extensions`...

It might be a good idea to have a section somewhere that lists them all for convenience.

It's likely that future spec iterations will add more (we've already been talking about things like `deprecated`, `version` & `author`), but I don't know if we should "landgrab" some others that we think we _might_ need in future.

If we do, we may unnecessarily limit what people can name their tokens and groups (e.g.  it seems `default` is already being used as a token name by some teams).

If we don't, we may may find it hard to add more properties in future spec versions.

So, how about this as an alternative: What if all spec properties started with a certain symbol. Taking a bit of inspiration from JSON schema, we could use a `$` prefix.

So, `value` becomes `$value`, `type` becomes `$type` and so on. An example file might then look like this:

```jsonc
{
  "$description": "The description of this group",
  "token name": {
    "$value": "100px",
    "$type": "dimension"
  },
  "another token": { 
    "$value": "#abcdef",
    "$type": "color"
  }
}
```

**Pros:**
- Easier to avoid naming clashes in the future
- Perhaps makes it easier to see which properties are spec things and which are author-defined token and group names

**Cons.**
- Might be a bit trickier to work with in code (e.g. in JavaScript you'd always need to access things like `token['$value']` intead of `token.value`)
[...]

PR #89, opened 2022-01-03
Moves group properties into a metadata property

c1rrus:
Updates the spec as per the discussion in issues #61 & #72 to move the `description` and `type` group properties into a `metadata` object to minimise naming clashes with group items now and in the future if more group properties are added.

### DIFF
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -6,18 +6,18 @@ A file may contain many tokens and they may be nested arbitrarily in groups like
 
 ```json
 {
-  "token uno": {
+  "token-uno": {
     "value": "token value 1"
   },
-  "token group": {
-    "token dos": {
+  "token-group": {
+    "token-dos": {
       "value": "token value 2"
     },
-    "nested token group": {
-      "token tres": {
+    "nested-token-group": {
+      "token-tres": {
         "value": "token value 3"
       },
-      "Token cuatro": {
+      "token-cuatro": {
         "value": "token value 4"
       }
     }
@@ -56,11 +56,37 @@ The names of items in a group are case sensitive. As per the guidance in the [de
   The format editors acknowledge existing best-practices for token naming, but place no direct constraints on naming via the specification.
 </p>
 
-## Additional group properties
+## Additional group metadata properties
+
+Groups may include additional metadata properties. To minimize naming clashes with items in the group, all metadata properties are nested inside an object whose key is `metadata`. `metadata` is therefore a reserved word within groups meaning that tokens or nested groups cannot be called `metadata`.
+
+<aside class="example">
+
+```json
+{
+  "space": {
+    "metadata": {
+      "description": "A description for the 'space' group",
+      "type": "dimension"
+    },
+    "small": {
+      "value": "0.5rem"
+    },
+    "medium": {
+      "value": "1rem"
+    },
+    "large": {
+      "value": "2rem"
+    }
+  }
+}
+```
+
+</aside>
 
 ### Description
 
-Groups may include an optional `description` property.
+Groups may include an optional `description` metadata property.
 
 For example:
 
@@ -69,9 +95,13 @@ For example:
 ```json
 {
   "brand": {
-    "description": "Design tokens from our brand guidelines",
+    "metadata": {
+      "description": "Design tokens from our brand guidelines"
+    },
     "color": {
-      "description": "Our brand's primary color palette",
+      "metadata": {
+        "description": "Our brand's primary color palette"
+      },
       "acid green": {
         "value": "#00ff66"
       },
@@ -91,17 +121,11 @@ Suggested ways tools may use this property are:
 - A GUI tool that lets users browse or select tokens could display this info alongside the corresponding group or as a tooltip
 - Export tools could output this as a source code comment
 
-<div class="issue" data-number="72">
-
-Groups may support additional properties like type and description. Should other properties be supported at the group level?
-
-</div>
-
 ### Type
 
-Groups may include an optional `type` property so a type property does not need to be manually added to every token. [See supported "Types"](#types) for more information.
+Groups may include an optional `type` metadata property so a type property does not need to be manually added to every token. [See supported "Types"](#types) for more information.
 
-If a group has a `type` property it acts as a default type for any tokens within the group, including ones in nested groups, that do not explicity declare a type via their own `type` property. For the full set of rules by which a design token's type is determined, please refer to the [design token type property chapter](#type-0).
+If a group has a `type` metadata property it acts as a default type for any tokens within the group, including ones in nested groups, that do not explicity declare a type via their own `type` property. For the full set of rules by which a design token's type is determined, please refer to the [design token type property chapter](#type-0).
 
 For example:
 
@@ -110,9 +134,10 @@ For example:
 ```json
 {
   "brand": {
-    "type": "color",
+    "metadata": {
+      "type": "color"
+    },
     "color": {
-      "description": "Our brand's primary color palette",
       "acid green": {
         "value": "#00ff66"
       },
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -58,7 +58,7 @@ The names of items in a group are case sensitive. As per the guidance in the [de
 
 ## Additional group metadata properties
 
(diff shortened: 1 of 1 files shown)


technical-reports/format/groups.md
```diff
@@ -6,18 +6,18 @@ A file may contain many tokens and they may be nested arbitrarily in groups like
...
   },
-  "token group": {
-    "token dos": {
+  "token-group": {
+    "token-dos": {
       "value": "token value 2"
     },
-    "nested token group": {
-      "token tres": {
+    "nested-token-group": {
```
dbanksdesign, line 16: Nice!

dbanksdesign (approved):
LGTM! :shipit:

ChucKN0risK:
I’ve reviewed this and have nothing to add/trust the others for the outcomes.

technical-reports/format/groups.md
```diff
@@ -110,9 +134,10 @@ For example:
 ```json
 {
   "brand": {
-    "type": "color",
+    "metadata": {
+      "type": "color"
```
drwpow, line 138: I like this. This allows someone to both set a default type for this group, as well as also create a `type` subgroup, which wouldn’t be out-of-the-question for typography tokens.

c1rrus: Good point! i hadn't actually considered that before.

In effect, this means that only `value` and `metadata` need to be reserved words. The reason is that `value` is the only required prop in a token and therefore the way in which parsers can differentiate between a group and a token.

I.e. if an object has a property called  `value`, it must be a token. If not, ituat be a group. Consequently, groups cannot contain items called `value`.

jorenbroekema: Yeah, name clashing crossed my mind when I read this spec today, I like this change as well!

drwpow: To expand on this: I’m currently working on a [parser for design tokens](https://github.com/drwpow/cobalt-ui) and the `type` key is a really important part of the schema, both for parsers as well as type systems like TypeScript to infer which is which. Having both a group and token possess a top-level `type: "color"` key would make everything exponentially harder/slower to reason about. So for that reason, I do like it being `metadata` on the group.

If group did have `type`, it’d make things easiest for parsers to be `type: "group"`. But that would probably cause confusion with the `metadata.type` proposal here, and overall that wouldn’t add _that_ much value for parsers as it can be inferred.

technical-reports/format/groups.md
```diff
@@ -56,11 +56,37 @@ The names of items in a group are case sensitive. As per the guidance in the [de
   The format editors acknowledge existing best-practices for token naming, but place no direct constraints on naming via the specification.
 </p>
 
-## Additional group properties
+## Additional group metadata properties
+
+Groups may include additional metadata properties. To minimize naming clashes with items in the group, all metadata properties are nested inside an object whose key is `metadata`. `metadata` is therefore a reserved word within groups meaning that tokens or nested groups cannot be called `metadata`.
```
kaelig, line 61:
```suggestion
Groups MAY include additional metadata properties. To minimize naming collisions with items in the group, all metadata properties are nested inside an object whose key is `metadata`. `metadata` is therefore a reserved word within groups, meaning that tokens and nested groups MUST NOT be called `metadata`.
```

technical-reports/format/groups.md
```diff
@@ -69,9 +95,13 @@ For example:
 ```json
 {
   "brand": {
-    "description": "Design tokens from our brand guidelines",
+    "metadata": {
+      "description": "Design tokens from our brand guidelines"
+    },
```
kaelig, line 100: The idea behind having `description` at the same level as `value` was to make it easily accessible, and ideally nudge authors toward documenting tokens.

Nesting the `description` field deeper _could_ have a negative impact on this.

Curious if that's something that was brought up when deciding about this change?

kaelig (commented):
A quick question and a suggestion!

technical-reports/format/groups.md
c1rrus: I don't recall that specific point being brought up, so thanks for raising it.

Bear in mind that this change only affects _groups_, not tokens. In tokens, `value`, `description` and other properties all remain top-level. That's because, as per the spec, no other properties are allowed within a token.

On groups, properties are used as the names for items within that group, so having "special" properties like `description` mixed in amongst them could be problematic if we ever added new group properties in future spec versions as there'd be the risk of naming collisions with item names people might be using. Moving all group properties under `metadata` avoids this.

As far as nudging folks to add descriptions to groups goes, not having `description` as a top level property on groups is of course less convenient to type. However, I doubt it's that much of a chore that it would put people of describing their group at all. With a bit of luck, IDEs may one day offer auto-completion (a la Intellisense) for editing DTCG files which would further aleviate the issue.

Arguably, having group `description` and other properties all collated into an object of their own may actually aid readability as they might otherwise be randomly mixed in amongst the group's item names.

I'm therefore still in favour of making this change. What do others think?

kaelig: Thank you for this great context, this makes more sense now 👍🏻 

And my apologies, I hadn't realized this was only impacting the schema of groups: my assumption was that if group descriptions were under `metadata.description`, for consistency they'd also have been nested in that location for tokens themselves.

This then raises the point: could this nesting inconsistency make the format harder to learn?

To illustrate my question, here's an explanation:

```json5
{
  amazingGroup: {
    // group description (nested under metadata object)
    metadata: { description: "my group" },
    // token description (direct child of the token)
    fantasticToken: { value: 15, description: "my token" },

    // expectation possibly set by group.metadata.description
    fantasticToken: {
      value: 15,
      metadata: { description: "my token" },
    }
  }
}
```

donnavitan: I like the idea of having consistent formatting whenever possible. And looking at this example now I can see the very subtle difference. Would having `metadata` on the group and the item create the same issue as having `description` at the top love?

drwpow: From a schema design perspective, it’s good to think in terms of what will break if a reserved word is added/removed/renamed. Namespacing (`metadata.*`) is a great way to mitigate against this.

Tokens are “closed,” meaning they only accept few reserved words (`type`, `value`, `description`). The potential for name changes won’t break much, as a user probably shouldn’t have any arbitrary keys on tokens. So in this scenario, `metadata` would just add length to everything when typing it out.

But groups are “open” in that a user may use any key they like. So any time a reserved word is added/removed/renamed, it may break all users’ schemas. Namespacing all reserved words into `metadata` (as the PR/Issue discuss) solves this.

All that to say it might _seem_ inconsistent on the surface to have `metadata` on one and not the other, but IMHO isn’t inconsistent in practice because tokens and groups behave differently and have different constraints, and it’s a reflection of that.

c1rrus: Thanks @drwpow , I couldn't have said it better myself :-)

c1rrus: For what it's worth, there was an earlier suggestion in #61 by @ivnmaksimovic to do the inverse: All group's top-level properties are "closed" - just like they are for tokens - but there is a special `tokens` property, which then contains just the child items. For example:

```json
{
  "group-1": {
    "description": "Description for group-1",
    "tokens": {
      "token-1": { "value": "..." },
      "token-2": { "value": "..." },
      "nested-group": {
        "tokens": {
          "deep-token-1": { "value": "..." }
        }
      }
    }
  }
}
```

I suppose you could argue it's more consistent from a stylistic point of view. But the trade-off is that it would require _every_ group to always contain a `tokens` property (regardless of whether or not is also uses any other special group properties like `description` or `type`). It also makes your JSON files nest more deeply which could be a bit inconvenient when hand-editing files.

There's pros and cons either way, but my personal preference is still to go ahead with `metadata` for groups only. I'm keen to hear the other format editor's views though.

c1rrus: I've added [a comment](https://github.com/design-tokens/community-group/issues/61#issuecomment-1011544318) in the "reserved names" issue (#61) mentioning that we decided at the last format editors' meeting to debate this further before deciding how to proceed.

As I've mentioned there, I'm now wondering if we should reconsider prefixing all "special" properites in our spec (both for groups and for tokens). That also avoids naming clashes but keeps the syntax for groups and tokens similar.

PR #90, opened 2022-01-04
Remove `Duration` from types to do list

aeons:
Duration is documented a couple of paragraphs above.


c1rrus (approved):
Thank you for your contribution @aeons !

PR #130, opened 2022-04-27
Add definition references for 'Design tool' and 'Translation tool'

kaelig:
This PR demonstrates how to link to definitions ([see documentation in ReSpec](https://respec.org/docs/#definitions-and-linking)).

If anyone finds this useful, please carry on adding more definition references!

It unlocks backlinks to the original definition for these terms. Here's how it looks like:

Screen Shot 2022-04-26 at 2 24 20 PM

Screen Shot 2022-04-26 at 2 24 08 PM


kaelig:
Merging this after a conversation with @valhead and @kevinmpowell

Issue #101, opened 2022-01-13
Gradient type feedback


c1rrus:
Is the specification for gradients that is being introduced by PR #86 fit for purpose? Does it need to also specify the type of gradient (.e.g linear, radial, concial, etc.)?

Please share your feedback, thoughts and ideas in this issue.

drwpow:
> Does it need to also specify the type of gradient (.e.g linear, radial, concial, etc.)? 

I do really like the proposal here, and I like the simplicity of having the gradient stops be universal, where you decide linear, radial, conical when generating code for a specific component (which in one part makes sense because a rotated gradient is still the same gradient).

But there also seems like an opportunity here to make `linear-gradient,` `radial-gradient`, and `conic-gradient` their own types to fill in the missing data (which exist as a concept both in CSS and Figma). Overall I fear that people may just abuse `metadata` on gradients here and put all the missing parts in there, at which point distinct types would be more useful. I could go either way—one gradient type with stops only, or distinct gradient types with complete properties.

TravisSpomer:
Omitting details like the type of gradient doesn't make much sense to me. A reasonable expectation is that for any given token, a design tool has everything it needs to display that gradient (like a Figma style), and a dev tool has everything it needs to turn that token definition into code. The current gradient spec doesn't work for either case. As it stands, I'm not really sure how I could use a gradient token for anything at all, without encoding all of the missing information as extensions.

At absolute minimum, I think that a gradient token needs to also include:

* The type of gradient
* The start and end points (0-1)

It would also be unfortunate if gradient stops were limited to specifying stops as a number 0-1. Other systems also allow stops to be defined as absolute units. For example, you might specify a gradient as going from black at 0, to transparent at 10px, to transparent at 1—that would get you a "poor man's shadow" that was always 10px wide no matter the size of the object you paint with it. That's admittedly a less common case, but it's something that I'm using today to support web and WinUI, and to continue supporting it I'd need to add an extension property like `"actuallyTheseAreTheRealGradientStops": [ 0, "10px", 1 ]`.

solemone:
Hello everyone, 

I have a question about the number of gradients. How about multiple gradients? Is the value then an array with arrays? Like like this:

```
{
  "blue-to-red": {
    "$type": "gradient",
    "$value": [
      [
        {
          "color": "rgba(0, 0, 255, 0.5)",
          "position": 0
        },
        {
          "color": "rgba(0, 0, 255, 1)",
          "position": 1
        }
[...]

PR #147, opened 2022-06-21
Color Module - First Draft

adekunleoduye:
### Overview 

This is the first draft of the Design Tokens Color Module.


technical-reports/color/color-type.md
```diff
@@ -1,22 +1,262 @@
...
+{
+  "Majestic magenta": {
+    "value": "#ff00ff",
+    "type": "color"
+  },
+  "Simple sage": {
+    "value": "#abcabc",
+    "type": "color"
+  }
+}
```
adekunleoduye, line 34: Update examples

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,262 @@
-# The color type
+# Color Type
 
-Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The value MUST be a string containing a hex triplet/quartet including the preceding `#` character. To support other color spaces, such as HSL, export tools SHOULD convert color tokens to the equivalent value as needed.
+Colors can be represented through various formats. For color, the type property must be set to the string “color”. For the value, the most common format to represent color through design tokens is a hex triplet. A hex triplet is a 6-digit, 24 bit, hexadecimal number that represents Red, Green, and Blue values as #RRGGBB.
```
kaelig, line 4: THis is a comment

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,257 @@
...
+  <tr>
+    <td>
+      <ul>
+        <li>Access to 50% more colors (P3 color space)</li>
+        <li>Colors more perceptually uniform</li>
+      </ul>
+    </td>
+    <td>
+      <ul>
+        <li>No fully supported (only safari)</li>
```
cgatian, line 247:
```suggestion
        <li>Not fully supported (only safari)</li>
```

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,377 @@
...
+
+There are many naming options when it comes to design tokens, especially color type tokens. We’ve identified two that seem to be most commonly implemented, **descriptive and numerical**.
+
+### Base Tokens
+
+For **Base tokens**, here’s how they may be represented in each version:
+
+#### Descriptive
+
+Descriptive names can be more emotional and human-friendly because they often relate to tangible things that people interact with, like grass or watermelons.
```
cgatian, line 56: I would note that using names like "grass" or "canary" may not make sense to someone whose first language is not English.  For projects that require engagement from designers and developers across languages this may not be a suitable choice.

technical-reports/color/color-type.md
cgatian: Just a nitpick. It would be nice if the Con sections mentioning the lack of browser support were consistent.

Something like 

> "Limited Browser Support (Safari)"

Consistent verbiage and browsers with support are listed.

technical-reports/color/token-naming.md
ayeshakmaz: Great point, thanks @cgatian!

technical-reports/color/index.html
```diff
@@ -80,9 +80,17 @@ <h1>Introduction</h1>
         <em>This section is non normative</em>
```
kaelig, line 80:
```suggestion
        <em>This section is non-normative.</em>
```

technical-reports/color/index.html
```diff
@@ -80,9 +80,17 @@ <h1>Introduction</h1>
...
-      <p>Work in progress.</p>
+      <section
+        data-include="./overview.md"
+        data-include-format="markdown"
+      ></section>
     </section>
 
+    <section
+      data-include="./token-naming.md"
+      data-include-format="markdown"
```
kaelig, line 91:
```suggestion
      data-include-format="markdown"
      class="informative"
```

^ Denotes this section is non-normative.

technical-reports/color/overview.md
```diff
@@ -0,0 +1,7 @@
+## Color module
```
kaelig, line 1: I think we may need to move this section directly into the `technical-reports/color/index.html` file like we do with the format: 

https://github.com/design-tokens/community-group/blob/d7fb89940415b8f61309b598527e55b6fea81e1d/technical-reports/format/index.html#L89-L100

And we can probably remove the `Color module` title.

technical-reports/color/color-type.md
```diff
@@ -12,11 +18,184 @@ For example, initially the color tokens MAY be defined as such:
...
+$translucent-shadow: ​hsl(153, 23%, 73%);
+```
+
+</aside>
+
+## Other value options
+
+### RGBA
+
+Formatted in R (red), G (green), B (blue) and (A) alpha. Red, green, and blue values can range from 0 to 255 and alpha values can range from 0 and 1 (i.e 0.5) or a percentage (i.e 50%) where 1 or %100 is full opacity.
```
kaelig, line 50:
```suggestion
Formatted in R (red), G (green), B (blue) and (A) alpha. Red, green, and blue values can range from `0` to `255`, and the alpha value ranges from `0` to `1` (such as `0.5`) or a percentage (such as `50%`) where `1` or `100%` is full opacity.
```

technical-reports/color/color-type.md
```diff
@@ -12,11 +18,184 @@ For example, initially the color tokens MAY be defined as such:
...
+
+</aside>
+
+### HSL
+
+Formatted in H (hue), S (saturation), L (lightness) and an optional (A) alpha. Hue values range from 0 to 360, saturation and lightness are percentage values that go from 0% to 100%, and alpha value can range from 0 and 1 (i.e 0.5) or a percentage (i.e 50%) where 1 or 100% is full opacity (which is the default value if a value isn’t provided).
+
+| Pros                                                  | Cons                                  |
+| ----------------------------------------------------- | ------------------------------------- |
+| It is easy to understand and compare to other formats | Not supported in all browsers (IE 11) |
```
kaelig, line 104: >Not supported in all browsers (IE 11)

Source?

technical-reports/color/color-type.md
```diff
@@ -12,11 +18,184 @@ For example, initially the color tokens MAY be defined as such:
...
+// colors-rgba.scss
+$majestic-magenta: hsl(300, 100%, 50%, 1);
+$simple-sage: hsl(100, 27%, 57%);
+```
+
+</aside>
+
+### Hex8
+
+Hex8 uses two extra digits, known as the alpha value, to change the transparency of the color. The format follows `#RRGGBBAA`. [Learn more about alpha values in hex.](https://www.digitalocean.com/community/tutorials/css-hex-code-colors-alpha-values#adding-an-alpha-value-to-css-hex-codes)
```
kaelig, line 143:
```suggestion
Hex8 uses two extra digits, known as the alpha value, to change the transparency of the color. The format follows `#RRGGBBAA`. [Learn more about alpha values in hex codes](https://www.digitalocean.com/community/tutorials/css-hex-code-colors-alpha-values#adding-an-alpha-value-to-css-hex-codes).
```

technical-reports/color/color-type.md
```diff
@@ -12,11 +18,184 @@ For example, initially the color tokens MAY be defined as such:
...
+// colors-rgba.scss
+$majestic-magenta: rgba(255, 0, 255, 0.5);
+$simple-sage: rgba(171, 202, 188, 0.5);
+```
+
+</aside>
+
+### LCH (Lightness Chroma Hue)
+
+Formatted in L (lightness), C (chroma), H (hue) and an optional (A) alpha. Hue values range from 0 to 360, saturation and lightness are percentage values that go from 0% to 100%, and alpha value can range from 0 and 1 (i.e., 0.5) or a percentage (i.e., 50%) where 1 or %100 is full opacity (which is the default value if a value isn’t provided).
```
kaelig, line 186:
```suggestion
Formatted in L (lightness), C (chroma), H (hue) and an optional (A) alpha. Hue ranges from `0` to `360`, saturation and lightness are percentage values that range from `0%` to `100%`, and the optional alpha value ranges from `0` and `1` (such as `0.5`) or as a percentage (such as `50%`) where `1` or `100%` is full opacity (which is the default value if a value isn’t provided).
```

technical-reports/color/color-type.md
```diff
@@ -12,11 +18,184 @@ For example, initially the color tokens MAY be defined as such:
...
+
+| Pros                                       | Cons                                                                        |
+| ------------------------------------------ | --------------------------------------------------------------------------- |
+| Access to 50% more colors (P3 color space) | Colors more perceptually uniform, so it can be harder to distinguish values |
+|                                            | Limited browser support (Safari only)                                       |
```
kaelig, line 191: >Limited browser support (Safari only)

Now supported in Chrome, Edge, Safari, and Firefox behind a flag.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,381 @@
...
+
+For **Base tokens**, here’s how they may be represented in each version:
+
+#### Descriptive
+
+Descriptive names can be more emotional and human-friendly because they often relate to tangible things that people interact with, like grass or watermelon.
+
+| Pros                                                                                                   | Cons                                                                                                                          |
+| ------------------------------------------------------------------------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------- |
+| Easier for some people to identify with, could be used to distinguish Brand colors from product colors | Harder to determine the scale of colors (i.e., which ones are lighter vs. darker, and how they pair well together)            |
```
kaelig, line 60:
```suggestion
| Easier for some people to identify with, could be used to distinguish brand colors from product colors | Harder to determine the scale of colors (for example: which ones are lighter vs. darker? How well do they pair together?)            |
```

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,381 @@
...
+// These can be used for background, border, or text colors
+$color-feedback-error: $color-red-600;
+$color-feedback-success: $color-green-400;
+```
+
+</aside>
+
+### Component Tokens
+
+Component specific names should start with the component that they support, and be located close to the component code. They commonly refer to alias tokens under the hood, and can be helpful when trying to use consistent styles across components while still maintaining separation of concerns.
```
kaelig, line 176:
```suggestion
Component-specific names should start with the component they support and be located close to the component code. They commonly refer to alias tokens and can be helpful when trying to use consistent styles across components while still maintaining separation of concerns.
```

I removed "under the hood", it's an idiom that international readers may not know the meaning of.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,381 @@
...
+  }
+}
+```
+
+</aside>
+
+<aside class="example">
+
+```jsx
+import {Theme} from "@design-system"
```
kaelig, line 369:
```suggestion
import {Theme} from "@design-system/components/Theme";
```

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,389 @@
...
+
+```json
+// tokens/color/alias.json
+{
+  "color": {
+    "$type": "color",
+    "text": {
+      "default": {
+        "$description": "Default text color",
+        "$value": {
```
kevinmpowell, line 206: @adekunleoduye is there any reason for single declarations we can't do:

```
...
"$value": "#111111"
...
```

The additional object with a single key of `"$hex"` seems unnecessary.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
-# The color type
+# Color type
 
-Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The value MUST be a string containing a hex triplet/quartet including the preceding `#` character. To support other color spaces, such as HSL, translation tools SHOULD convert color tokens to the equivalent value as needed.
+## Format
 
-For example, initially the color tokens MAY be defined as such:
+Colors can be represented through various formats. For color tokens, the `$type` property MUST be set to the string `color`. The `$value` property can then be used to specify the details of the color, including color space, alpha settings, and more. The `$value` object contains the following properties:
+
+- `$hex` (required): the hex color to use as the default or a guaranteed fallback
```
drwpow, line 7: Suggestion: `$hex`, `$colorSpace`, `$name` etc should probably drop the `$` prefix, right? For all other tokens, `$value` objects contain normal non-prefixed keys. Which feels more expected because outside the token, `$value` is a special reserved word that marks any object as a token. But inside the `$value`, a strict schema is enforced where arbitrary properties are no longer allowed (so no benefit to prefixing keys).

Apologies if there’s a discussion I missed or if this suggestion isn’t helpful.

technical-reports/color/token-naming.md
kaelig: My understanding was that this notation was still supported?

technical-reports/color/gradient.md
c1rrus, line 1: I suggest splitting this out into a separate PR for now.

As it stands, this has got quite a few deviations from [how the gradient type is currently defined in the format](https://tr.designtokens.org/format/#gradient):

* The format does not use `$` prefixes for the value's inner properties and, as per the comments from @drwpow and myself on the updated color type, I think we should stick with that. (or, if we didn't, we'd need to change all the other composite types to be consistent)
* The format does not include a `style` property, though several folks have suggested something along those lines in #101. I think that's probably a worthwhile addition, but there are other interesting proprosals in that thread too, which I think we ought to consider holistically.
* The format does not have an `alpha` property on the stops and, given that the color values can have an alpha channel, it feels kinda redundant. Besides, what would be the expected behaviour if both have been specified but have conflicting values?

I presume the intent of this section is really just to show that the new, expanded color value syntax defined in the previous chapter can be used anywhere else that color values are permitted.

The `value` of a gradient stop is one case of that, but there are others too:

- the `color` sub-value of a shadow
- the `color` sub-value of a border

So, perhaps this section could be reworked into something that just states that the new object-style color values can be used throughout the spec and include an example of each of those things?

c1rrus (changes requested):
Sorry it's taken me so long to review this. Great work putting all of this together!

However, I think there's still work to be done to better align this with the format spec.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
...
+- `$colorSpace` (optional): An object detailing an alternative color space to be used to interpret the color, if supported
+
+The `$colorSpace` object has the following properties:
+
+- `$name` (required): the name of the color space (either rgb, srgb, hsl, or lch)
```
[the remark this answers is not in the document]
phun-ky: I agree, since `colorSpace` is optional, it can have a more distinct nomenclature instead of `name`

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,389 @@
+# Token naming
+
+## Categorization
+
+There are 3 main categories of design tokens that we will continue to reference in this specification.
+
+### Base
+
+Base tokens are the lowest level tokens and typically consist of a name and [hexadecimal](https://www.w3.org/TR/css-color-4/#hex-notation) value pair.
```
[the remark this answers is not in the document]
phun-ky: If the naming convention for color tokens are in the same format as the others (most likely?), why don't we reference a global naming convention here?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
...
+- `$components` (required): the non-alpha pieces of the color, listed as an array of floating-point numbers or integers
```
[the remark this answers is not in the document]
drwpow: Oh yes this **definitely needs clarification!** Because no, values won’t always be between `0-1` if we follow [CSS Color Module 4](https://www.w3.org/TR/css-color-4/). There are 2 reasons:

1. **Degrees**. As in the case of HSL, the range is 0°–360°. This could probably be normalized to `1` but I think most would agree that fails the goal of this spec being human writable since that’s nonstandard
2. **Device-independent colorspaces.** In the case of colorspaces like LAB/LCH and OKLAB/OKLCH, they describe many colors that are outside the range of any device existing today, and so there’s no “maximum.”

For the latter, CSS Color Module 4 still does define allowable ranges which are still far beyond what most devices can support; I believe they just put reasonable caps on values that decades of hardware improvements still wouldn’t catch up to. Here are some of the values:

- Lab: L [0 … 100], a [-125 … 125], b [-125 … 125]
- Lch: L [0 … 100], c [0 … 150], h [0 … 360]
- Oklab: L [0 … 1], a [-0.4 … 0.4], b [-0.4 … 0.4]
- Oklab: L [0 … 1], c [0 … 0.4], h [0 … 360]
- XYZ: 🤔 for some reason this is normalized [0 … 1] but not sure why

The general approach of “match CSS Color Module 4 as much as possible” has been discussed in meetings but I think it would be good to explicate this as much as possible. As far as I know, the color proposal wouldn’t deviate from Color Module 4 in any way. cc @resource11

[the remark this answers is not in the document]
drwpow: Also `alpha` in all cases is always normalized [0 … 1], and is optional (default `1` if omitted).

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
...
+- `$alpha` (optional): the alpha component of the color, listed as a floating-point number integer in the range of 0 to 1. If omitted, color is assumed to be fully opaque
+
+## Hex - required support
+
+For the color value, the required fallback format to represent color through design tokens is a hex or hex8 value. A hex triplet is a 6-digit, 24 bit, hexadecimal number that represents Red, Green, and Blue values as `#RRGGBB`. An eight-character hex will include the alpha value in the last 2 characters. If no alpha value is specified, it is assumed the color if fully opaque.
```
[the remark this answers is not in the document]
drwpow: Agreed. Hex values have started to be used for Display P3, but the intent is limited to sRGB.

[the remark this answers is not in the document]
Shrinks99:
```suggestion
For the color value, the required fallback format to represent color through design tokens is a hex or hex8 value. A hex triplet is a 6-digit, 24 bit, hexadecimal number that represents Red, Green, and Blue values as `#RRGGBB`. An eight-character hex will include the alpha value in the last 2 characters. If no alpha value is specified, it is assumed the color if fully opaque. The sRGB colorspace MUST be used for interpreting fallback hex values.
```

technical-reports/color/color-type.md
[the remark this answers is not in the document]
phun-ky: @drwpow yeah, if no color space is omitted, and no other color function is used, opacity/alpha channel (rgba) should always default to normalized 0-1

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
-# The color type
+# Color type
 
-Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The value MUST be a string containing a hex triplet/quartet including the preceding `#` character. To support other color spaces, such as HSL, translation tools SHOULD convert color tokens to the equivalent value as needed.
+## Format
 
-For example, initially the color tokens MAY be defined as such:
+Colors can be represented through various formats. For color tokens, the `$type` property MUST be set to the string `color`. The `$value` property can then be used to specify the details of the color, including color space, alpha settings, and more. The `$value` object contains the following properties:
```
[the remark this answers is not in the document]
drwpow: Note: resolving this comment (and others) that are old. Going to make new comments to reflect recent color group meetings and internal feedback shared among DTCG editors. Resolving right now means “feedback heard and accepted; ALSO there will be imminent changes to this PR.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,229 @@
...
+
+- `$hex` (required): the hex color to use as the default or a guaranteed fallback
+- `$colorSpace` (optional): An object detailing an alternative color space to be used to interpret the color, if supported
+
+The `$colorSpace` object has the following properties:
```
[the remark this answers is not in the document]
drwpow: Question: should we explain why it’s an array, and why the order matters (following CSS Color 4 spec)? Or is that too much information?

technical-reports/color/gradient.md
```diff
@@ -0,0 +1,134 @@
+# Gradient
```
drwpow, line 1: - [ ] Remove this file as it’s a separate proposal (but this can be a new PR if we want to review)
- [ ] Changes instead should go in `composite-types.md`—updating all those examples to reference valid colors

technical-reports/color/index.html
```diff
@@ -77,16 +79,29 @@
       <h1>Introduction</h1>
 
       <p>
-        <em>This section is non normative</em>
+        <em>This section is non-normative.</em>
       </p>
 
-      <p>Work in progress.</p>
+      <section
+        data-include="./overview.md"
```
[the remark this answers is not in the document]
drwpow: - [ ] (This comment is still valid and needs addressing)

technical-reports/color/index.html
```diff
@@ -77,16 +79,29 @@
       <h1>Introduction</h1>
 
       <p>
```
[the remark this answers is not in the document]
drwpow: - [ ] (This comment still needs to be addressed—this info should all go into `technical-reports/format/types.md`)

technical-reports/format/groups.md
```diff
@@ -79,7 +79,8 @@ Group keys without a dollar sign (`$`) prefix denote:
     "Group of tokens": {
       "$description": "This is an example of a group containing a single token",
       "Token name": {
-        "$value": "#000000"
+        "$value": "#000000",
```
drwpow, line 82: - [ ] Here and blow, for `$type: "color"` tokens, update examples to match current updated format

technical-reports/format/types.md
```diff
@@ -12,7 +12,7 @@ If an explicit type is set, but the value does not match the expected syntax the
 
 ## Color
 
-Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The value MUST be a string containing a hex triplet/quartet including the preceding `#` character. To support other color spaces, such as HSL, translation tools SHOULD convert color tokens to the equivalent value as needed.
+Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The `$value` property MUST be a string containing a [hex triplet/quartet](https://www.w3.org/TR/css-color-4/#typedef-hex-color) including the preceding `#` character. To support other color spaces, such as HSL, translation tools SHOULD convert color tokens to the equivalent value as needed.
```
drwpow, line 15: - [ ] (see [@c1rrus’ other comment](https://github.com/design-tokens/community-group/pull/147/files#r1392453903) about moving changes here)

technical-reports/index.html
```diff
@@ -65,7 +66,7 @@ <h1>Modules</h1>
         <li>
           <a href="./format/">Format</a>
         </li>
-        <li>Color (coming soon)</li>
+        <li><a href="./color/">Color</a></li>
         <li>Animations (coming soon)</li>
```
drwpow, line 70: Unknown how best to update this (related to [@c1rrus’ comment](https://github.com/design-tokens/community-group/pull/147/files#r1392453903)), but perhaps we consolidate everything into `format/types.md`?

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,389 @@
+# Token naming
```
drwpow, line 1: - [ ] Remove this document? Since we’re updating the format itself, and not providing recommendations (this starts to conflict with other upcoming proposals like modes), we could always split out ideas here into separate proposals or separate conversations to be discussed apart from the color changes.

drwpow:
Been thinking more about the minor difference between `channels: [x, x, x]` vs `components: [x, x, x]` and I don’t know where/when either originated. But re-reading the Color Module 4 (and 5) specification, as well as various Wikipedia pages for color science, and `components` does actually appear more frequently than `channels`.

Reading further, there are some minor nuances between when to use each term, but I think for our purposes / for the purposes of the average user they are likely moot (as hinted at by CSS Color Module 4’s wording “components are sometimes called ’channels,’” also not concerned with distinguishing between the two. For the frontenders, it almost feels like the distinction between “HTML attributes“ vs “HTML properties.”

I think either term would work, but just based on my reading, seeing “component” used more commonly, I’m 1% leaning toward that term, personally. I think in hindsight latching onto “channel” was just from my time spent in Photoshop.

nesquarx:
Given the mathematical nature of some of the colour spaces (LAB for
example), they are not really representative of colour channels all the
time. Component feels more inclusive.

adekunleoduye:
Closing since we have an updated version in #266.

PR #151, opened 2022-07-01
refactor(part 6.2.2): fix overflow-x caused by a non responsive image

lauthieb:
Hello,

_First of all, a huge thank you to all the people who have contributed so far to this wonderful initiative.
Being passionate about UI and standardization, I loved reading the second draft and I want to be part of the contributors by proposing some changes that I hope you will like._

In this PR, I propose you to fix a UI bug in the format module documentation.

## Bug description

Because of the image in part 6.2.2 "GUI Tools" was declared in Markdown format, it generated an image with a fixed width of `426px`. Setting a `width: 100%` and `max-width: 426px` solves this problem for mobile screens.

## Before

CleanShot 2022-07-01 at 02 06 19@2x

## After

CleanShot 2022-07-01 at 02 05 58@2x

Please, don't hesitate if you have any remarks.

Thanks in advance for your review!

### DIFF
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -232,7 +232,10 @@ For example:
 
 Tools that let users pick or edit tokens via a GUI MAY use the grouping structure to display a suitable form of progressive disclosure, such as a collapsible tree view.
 
-<img src="./group-progressive-disclosure.png" alt="Progressive disclosure groups" style="width:100%; height: auto;" />
+<figure id="figure-group-progressive-disclosure">
+  <img src="./group-progressive-disclosure.png" alt="" />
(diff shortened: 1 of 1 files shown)


kaelig:
Merci Laurent ! After looking into it, I found there are styles available for images to adapt correctly to viewport sizes, as seen in https://respec.org/docs/#figures-table-of-figure, so I went ahead and updated the PR to use that method in [`b9aaa98` (#151)](https://github.com/design-tokens/community-group/pull/151/commits/b9aaa985a6a5101b6d1068139eb75df6bc2653bb)

lauthieb:
Hi @kaelig, de rien ! :)
Oh yes, you're right, thanks for this 🙏

Issue #110, opened 2022-02-03
Add a `$private` property for tokens


c1rrus:
Should our spec add an optional `$private` property, along similar lines to [the one proposed for StyleDictionary](https://github.com/amzn/style-dictionary/pull/704) by @silversonicaxel? It would instruct tools not to display or export a token by default (though tools may provide an option for users to override that and still see / export private tokens if they wish). Its value is a boolean. `true` makes the token private, `false` makes it (explicitly) public. If there is no `$private` property, the token is public.

As per the thread on the StyleDictionary issue, I think this could be convenient way for teams to exclude (or flag) tokens that they don't consider part of their public API from code, design tools, styelguides, etc. 

For example, the following DTCG token file:

```jsonc
{
  "red": {
    "$type": "color",
    "$value": "#ff0000"
  },
  "green": {
    "$type": "color",
    "$value": "#00ff00",
    "$private": true
  },
  "blue": {
    "$type": "color",
    "$value": "#0000ff",
    "$private": false
  }
}
```

...might cause a SASS export tool to output code like this:

```scss
$red: #ff0000;
$blue: #0000ff;
```
(Note how the "green" token is omitted because it is private)

Furthermore, if a token is an alias to a private token, then the dereferenced value must be output. E.g.:

```json
{
  "red": {
    "$type": "color",
    "$value": "#ff0000"
  },
  "blue": {
    "$type": "color",
    "$value": "#00ff00",
    "$private": true
  },

  "danger-color": {
    "$value": "{red}"
  },
  "link-color": {
    "$value": "{blue}"
  }
}
```

...might be exported to SASS like so:

```
$red: #ff0000;
$danger-color: $red;
$link-color: #0000ff;
```
(Note how `$link-color` has the dereferenced value, because the "blue" token is private)

Finally, I'd suggest that this property should be inheritable from groups, just like `$type` is. Thiis would make it easier to make all tokens within a group private. For example:

```json
{
  "color": {
    "$type": "color",
    "$private": true,

    "black": {
      "$value": "#000000"
    },
    "white": {
      "$value": "#ffffff"
    }
  }
}
```

is equivalent to:

```json
{
  "color": {
    "$type": "color",

    "black": {
      "$value": "#000000",
      "$private": true
    },
    "white": {
      "$value": "#ffffff",
      "$private": true
    }
  }
}
```

What do you think?

jeromefarnum:
[...]

PR #172, opened 2022-09-26
Adds new token property (`$private`)

TremayneChrist:
Adds support for private tokens, to make defining tokens easier, without bloating any token outputs.

A simple use case could be using internal color palettes to create gradients, for example:

```json
{
  "color": {
    "$private": true,
    "red": {
      "$value": "red"
    },
    "orange": {
      "$value": "orange"
    },
    "yellow": {
      "$value": "yellow"
    },
    "green": {
      "$value": "green"
    },
    "blue": {
      "$value": "blue"
    },
    "indigo": {
      "$value": "indigo"
    },
    "violet": {
      "$value": "violet"
    }
  },
  "gradient": {
    "$type": "gradient",
    "rainbow": {
      "$value": [
        {
          "color": "{color.red}",
          "position": "0"
        },
        {
          "color": "{color.orange}",
          "position": "0.166"
        },
        {
          "color": "{color.yellow}",
          "position": "0.333"
        },
        {
          "color": "{color.green}",
          "position": "0.5"
        },
        {
          "color": "{color.blue}",
          "position": "0.666"
        },
        {
          "color": "{color.indigo}",
          "position": "0.833"
        },
        {
          "color": "{color.violet}",
          "position": "1"
        }
      ]
    }
  }
}
```

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -130,6 +130,36 @@ For example:
 
 </aside>
 
+## Private
+
+Marks a token as private.
+
+Private tokens can be used to create aliases and other token combinations, reducing duplication and simplifying token outputs. Tools can resolve private tokens internally, but should not look to make them publicly available by default.
+
+A token's visibility can be specified by the optional `$private` property. If the `$private` property is not set to `true` on a token, then the token's visibility is visible.
+
+The `$private` property can be set on different levels:
+
+- at the group level
+- at the token level
+
+The value of the `$private` property MUST be a boolean, whose value is either `true` or `false`.
+
+For example:
+
+<aside class="example">
+
+```json
+{
+  "Button background": {
+    "$value": "#777777",
+    "$private": true
+  }
+}
+```
+
+</aside>
+
 ## Extensions
 
 The optional **`$extensions`** property is an object where tools MAY add proprietary, user-, team- or vendor-specific data to a design token. When doing so, each tool MUST use a vendor-specific key whose value MAY be any valid JSON data.


technical-reports/format/design-token.md
```diff
@@ -130,6 +130,36 @@ For example:
 
 </aside>
 
+## Private
+
+Marks a token as private.
+
+Private tokens can be used to create aliases and other token combinations, reducing duplication and simplifying token outputs. Tools can resolve private tokens internally, but should not look to make them publicly available by default.
```
romainmenke, line 137: This effectively makes this a write-only storage.

Once a designer marks a token as private it would become invisible to them because their tool would need to honour the flag.

It might be more interesting to have "filter" tools that teams can use to convert one or more "source" token files into a "filtered" final token file.

romainmenke: 🤔 Tools could be required to have a setting to show or hide `private` tokens.

TremayneChrist: It depends on how you see these files being used. A token editor tool, could consume the input JSON and would be able to see all tokens and allow you to see/modify the private flag.

A translator on the other hand, would resolve private tokens, but would not export them to any output files (e.g. CSS, XML, JSON etc.)

I feel designers and developers should be consuming a processed output and see the input JSON as the source of truth, which only a certain set of designers modify. This could be an alternative approach to filter tools.

TremayneChrist: So this wording would need to change to be clearer around different types of tooling.

romainmenke: Would it then not be better to have a token management app/service which stores this information under `$extension`.

I don't think this requires a spec change.

--------

> which only a certain set of designers modify.

I don't think it makes sense to encode a hierarchy of control over tokens in the spec.

I do think it can be useful to have a filter to reduce clutter.

TremayneChrist: > I don't think it makes sense to encode a hierarchy of control over tokens in the spec. (what if tokens are created by developers with dev tools in a certain org?)

Yeah I'm not saying this should be defined in the spec, it needs to be flexible for anyone to use. A token change could have disarsterous effects, though. So, this would no doubt be limited to certain editors who are working on large projects. I agree with you, the spec should definitely stay clear of all of this, I was only giving an example. My main point is that I feel designers and developers should consume the same data, albeit in different formats. Token editors, whoever they may be, would edit the input JSON - either directly, or via other tooling, and they would be able to see all data, including the visibility of a token.

Adding private token support, allows structuring and reusablity of internal values - it effectively makes things a lot cleaner for token editors. Further filtering could be beneficial for other teams, however, I see private tokens as a separate thing, which is beneficial to the spec. That's if, we have agreement on both designers and developers consuming the same (processed) data?

romainmenke: My main hesitation with this is that I don't think `token editors` can ever be a well defined role with tools exclusively for them.

I don't think you can write a spec for `$private` because it's not realistic to define which tools should show or hide the tokens.

Ultimately anyone with a plain text editor can still open the tokens file and remove the `$private` flag.

I just don't think this is the right API/abstraction.
But I am just one person, with one persons point of view :)

TremayneChrist: Writing a spec is extremely difficult and you need to be hesitant with any new request - I fully support that! :)

I guess I'm trying to understand how you would see all of this being used in the real would. Taking 'token editors' out of the equation, would you see the `tokens.json` as input and translated files as output? If so, then `$private` should make sense, as the properties configure the output. If not, I would like to know more about how and where this spec would be implemented.

FYI: I'm fully up for being more involved around all of this.

romainmenke: I'm a community contributor and translation tool author (not an editor), so my opinions are my own and my concerns are influenced by my role as a translation tool author.

Personally I prefer to trust and empower downstream consumers.
Not hiding anything creates transparency which will foster trust and open communication between teams and team members.

-------

> Adds support for private tokens, to make defining tokens easier, without bloating any token outputs.

> would you see the tokens.json as input and translated files as output? If so, then $private should make sense, as the properties configure the output. If not, I would like to know more about how and where this spec would be implemented.

Is the issue you are trying to address that translation tools mostly generate code and do this for all tokens.

A tokens file following the design tokens methodology will have a lot of tokens intended for aliasing.
A translation tool for CSS will generate `--foo: #000;` declarations for each without.

The end result has a lot of bloat because the translation tool doesn't validate if a token value is used.

romainmenke: Also see https://github.com/design-tokens/community-group/issues/110

TremayneChrist: The issue that I'm facing is that I want to use **_tokens_** to create **tokens**. However, I don't want the **_tokens_** I used to create the **tokens** to be pomoted for use - They are used for computing other tokens, and I want to pass this information to translators so that they can choose how to handle them. An option could be passed to a translator to include them in the output. It's a way to structure `tokens.json` and inform translators which tokens are relevant and should be exported by default. It may only be a single token that is marked as private, but is used to create many other token values.

The example in the description is probably not the best. It hides colors from designers, which is not ideal.

TremayneChrist: > Double checking because private has a very strong association with access control and that is a different thing.

`$hidden`?

romainmenke: Classes in certain programming languages also have to concepts of `abstract` and `final`. I think that has some overlap.

`abstract` -> only intended to be used to construct other tokens
`final` -> never intended to be used to construct other tokens

Tokens with `$abstract` :
- can only be used via an alias
- never possible to access their value directly

Tokens with `$final` :
- can never be aliased

_You don't need both concepts._

kaelig:
Thank you so much for this contribution!

**Note: this is my opinion and this isn't a DTCG decision as I haven't discussed it with other editors:** I'd be in favor of shelving this for now, and see if this becomes a strong need that would make sense to provide as part of the spec.

My assumption is that people who want "private" tokens could use a prefix like underscore (`_myPrivateToken`) or a metadata extension field, so that the appropriate part of their toolchain knows when to omit a token in the output.

technical-reports/format/design-token.md
TremayneChrist: After thinking about this a bit more, I don't think we need to go that far, we're not designing a new programming language here - it's essentially a configuration file. `$abstract` and `$final` feel a bit too much.

I agree that `$private` is the wrong name, however, and I would suggest it to be more like a visibility state, hidden/visible. Token editing tools would be able to see all tokens and modify the visibility state for translators.

`$hidden` seems like a good option.

kaelig:
I totally see the need for "omit from output" functionality, but not sure it should be in the core grammar of the spec right now.

For example, you may want to output a certain token for a particular case, and omit it in another (iOS vs Android).
We haven't tackled multi-file token sets yet, and I have the feeling this is when private/public tokens will become very relevant (in a similar fashion as ES imports).

kevinmpowell:
I'm in agreement with @kaelig on this. I don't think this is something needed right now.

PR #180, opened 2022-11-13
📝 Fix small typos

PavelLaptev:
Just a few typos to fix

### DIFF
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -27,7 +27,7 @@ A file MAY contain many tokens and they MAY be nested arbitrarily in groups like
 
 </aside>
 
-  The names of the groups leading to a given token (including that token's name) are that token's _path_, which is a computed property. **It is not specified in the file**, but parsers that conform to this spec MUST be able to expose the path of a token. The above example, therefore, defines 4 design tokens with the following properties:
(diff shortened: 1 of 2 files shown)


technical-reports/format/groups.md
```diff
@@ -27,7 +27,7 @@ A file MAY contain many tokens and they MAY be nested arbitrarily in groups like
 
 </aside>
 
-The names of the groups leading to a given token (including that token's name) are that token's _path_, which is a computed property. **It is not specified in the file**, but parsers that conform to this spec MUST be able to expose the path of a token. The above example, therefore, defines 4 design tokens with the following properties:
+  The names of the groups leading to a given token (including that token's name) are that token's _path_, which is a computed property. **It is not specified in the file**, but parsers that conform to this spec MUST be able to expose the path of a token. The above example, therefore, defines 4 design tokens with the following properties:
```
kaelig, line 30: Did you mean to indent this paragraph?

```suggestion
The names of the groups leading to a given token (including that token's name) are that token's _path_, which is a computed property. **It is not specified in the file**, but parsers that conform to this spec MUST be able to expose the path of a token. The above example, therefore, defines 4 design tokens with the following properties:
```

kaelig (approved):
Great catches, thank you so much! Just curious about an indentation change.

technical-reports/format/groups.md
PavelLaptev: @kaelig yes, since it's a MD file it doesn't do anything and disappears after conversion to HTML.
Below, a screenshot from the last published draft.

image

kaelig:
Thank you @PavelLaptev, your updates are now live!

Issue #53, opened 2021-09-07
[Format] Type: font name


kaelig:
A naive approach like the one below may be appropriate for the first stage of the specification, but this may be more complicated than it seems due to platform/OS/browser restrictions.

Please share feedback, ideas and concerns in the issue below so we can assess the best course of action.

---

### Font name

Represents a font name or an array of font names (ordered from most to least preferred). The type property must be set to the string “font”. The value must either be a string value containing a single font name or an array of strings, each being a single font name. For example:

```json
{
  "Primary font": {
    "value": "Comic Sans MS",
    "type": "font"
  },
  "Body font": {
    "value": ["Helvetica", "Arial"],
    "type": "font"
  }
}
```

lucastobrazil:
Just sharing my current setup. We use tokens for both Figma and Web platforms. Figma requires the "font name" to be specified so it can also use the `loadFontAsync()` function, so we store both the name and the CSS string for `font-family`. 

I'd imagine this could be more explicit by maybe including `vendor` in there, so it's clearer what it's used for, but putting it out there for the discussion. 

```ts
type FontFamilyData = {
  fontName: string; // For Figma
  fontFamily: string; // For Web
};

interface FontFamilyToken extends BaseToken {
  name: string;
  description: string;
  data: FontFamilyData;
}

const FontFamilies: { [x: string]: FontFamilyToken } = {
  sans: {
    name: "sans",
    description: "Sans-serif font used for all UI text.",
    data: {
      fontName: "SF Pro Text",
      fontFamily:
        '-apple-system, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif',
    },
  },
  sansDisplay: {
    name: "sans-display",
    description: "Display variant of sans-serif font.",
    data: {
      fontName: "SF Pro Display",
      fontFamily:
        '-apple-system-headline, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Fira Sans", "Droid Sans", "Helvetica Neue", sans-serif',
    },
  },
};

export default FontFamilies;

```

c1rrus:
Just posting a small update since things have moved on a bit since this issue was opened. The latest draft spec has:

* Added a [`fontWeight` type](https://design-tokens.github.io/community-group/format/#font-weight)
[...]

PR #181, opened 2022-11-16
Promote usage of a generic family name in the array example

kizu:
While the issue of typing the font-family is still open (https://github.com/design-tokens/community-group/issues/53), I think it could still be a good idea to list a generic font-family name in the array example as a generally good practice, for example, see [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/font-family):

> You should always include at least one generic family name in a font-family list, since there's no guarantee that any given font is available.

### DIFF
--- a/technical-reports/format/types.md
+++ b/technical-reports/format/types.md
@@ -101,7 +101,7 @@ For example:
     "$type": "fontFamily"
   },
   "Body font": {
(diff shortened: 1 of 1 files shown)


c1rrus (approved):
Good idea. Thank you for this contribution!

kevinmpowell (approved):

PR #182, opened 2022-11-16
Fix the color alpha syntax to be 50%

kizu:
Noticed an issue with the [Example 15](https://design-tokens.github.io/community-group/format/#example-15) and [https://design-tokens.github.io/community-group/format/#example-16](Example16): Per the spec text, the 24 bit color value is expected to be transformed into a color with the alpha value of `50%`, while `88` would be actually around `53%`, being a common misconception when writing the component in 8bit format.

This PR changes any mentioned `#00000088` colors to `#00000080` to match the intent and fix the examples.

### DIFF
--- a/technical-reports/color/color-type.md
+++ b/technical-reports/color/color-type.md
@@ -13,7 +13,7 @@ For example, initially the color tokens MAY be defined as such:
     "$type": "color"
   },
   "Translucent shadow": {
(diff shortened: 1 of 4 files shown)


c1rrus (approved):
Good spot! Thank you so much for contributing these corrections!

PR #183, opened 2022-11-17
Clarify that 0 needs a unit as well.

lukasoppermann:
This question has come up numerous times at my work.

In css `0` needs no unit as units are multiplied with the number and `0*px === 0`.

However the way I understand the specs this is not true atm for design tokens, correct?

In this case it would be great to clarify this.

### DIFF
--- a/technical-reports/format/types.md
+++ b/technical-reports/format/types.md
@@ -66,7 +66,11 @@ For example:
 
 ```json
 {
-  "spacingStack1X": {
+  "spacing-stack-0": {
+    "$value": "0rem",
+    "$type": "dimension"
+  },
+  "spacing-stack-1": {
     "$value": "0.25rem",
     "$type": "dimension"
   }
--- a/technical-reports/format/types.md
+++ b/technical-reports/format/types.md
@@ -58,7 +58,7 @@ $translucent-shadow: hsla(300, 100%, 50%, 0.5);
 
 ## Dimension
 
-Represents an amount of distance in a single dimension in the UI, such as a position, width, height, radius, or thickness. The `$type` property MUST be set to the string `dimension`. The value must be a string containing a number (either integer or floating-point) followed by either a "px" or "rem" unit (future spec iterations may add support for additional units).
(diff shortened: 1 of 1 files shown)


c1rrus (approved):
Thanks for adding this!

You're quite right, as its specced, the dimension values must have a unit even when they are zero.

It's worth noting that a translation tool would be welcome to omit the unit in the output if it makes sense to do so for the output language.

For example `"$value": "0rem"` could be exported as just `0` in CSS if you wanted.

technical-reports/format/types.md
```diff
@@ -58,7 +58,7 @@ $translucent-shadow: hsla(300, 100%, 50%, 0.5);
 
 ## Dimension
 
-Represents an amount of distance in a single dimension in the UI, such as a position, width, height, radius, or thickness. The `$type` property MUST be set to the string `dimension`. The value must be a string containing a number (either integer or floating-point) followed by either a "px" or "rem" unit (future spec iterations may add support for additional units).
+Represents an amount of distance in a single dimension in the UI, such as a position, width, height, radius, or thickness. The `$type` property MUST be set to the string `dimension`. The value must be a string containing a number (either integer or floating-point) followed by either a "px" or "rem" unit (future spec iterations may add support for additional units). This includes `0` which also MUST be followed by either a "px" or "rem" unit.
```
aaronccasanova, line 61: This is helpful context and may be worth including in the following code block/example. Thoughts?

```jsonc
{
  "spacing-0": {
    "$value": "0rem",
    "$type": "dimension"
  },
  "spacing-1": {
    "$value": "0.25rem",
    "$type": "dimension"
  },
  // etc.
}
```

lukasoppermann: You mean to add the `0rem` to the existing example? I am fine either way, don't know if it is needed, but I think it for sure does not hurt.

aaronccasanova: Yes, updating the existing example 👍 Every "MUST" condition thus far is demonstrated and think this was a great callout that might otherwise be overlooked. Like you mentioned it doesn't hurt to add so I'll trust your judgement on whether or not to include!

aaronccasanova (approved):

kevinmpowell (approved):
@c1rrus this one looks good to merge.

Issue #139, opened 2022-06-19
$type should be a required property


reblim:
[3.2 (Design) Token Properties](https://second-editors-draft.tr.designtokens.org/format/#design-token-properties)

A concern I have with the current specification is that we are not making the "$type" property required. I believe that both the "$type" and "$value" properties should always be required to avoid tools from having to guess a type based on the group type, or even higher scopes.

Nevertheless, in order for this to work well, we must define our core types more precisely and create a complete list of all possible "primitive" token types. In addition, we could have a "default" type that acts like "any" in TypeScript, to be used if a defined type does not provide enough context for a token.

romainmenke:
> A concern I have with the current specification is that we are not making the "$type" property required. I believe that both the "$type" and "$value" properties should always be required to avoid tools from having to guess a type based on the group type, or even higher scopes.

I fully agree with this.
Even for `string` and `number` the `$type` field should be set in my opinion.
see : https://github.com/design-tokens/community-group/issues/120

Composite types should be typed groups.
see : https://github.com/design-tokens/community-group/issues/126

----

> Nevertheless, in order for this to work well, we must define our core types more precisely and create a complete list of all possible "primitive" token types. In addition, we could have a "default" type that acts like "any" in TypeScript, to be used if a defined type does not provide enough context for a token.

This I do not see the purpose of, but I might be missing the point :)

Tools would be unable to use this token as they do not know its type.

So a `default` or `any` type would create "write only" tokens.
You could store them, but you would be unable to read them.

reblim:
> > This I do not see the purpose of, but I might be missing the point :)

Tools would be unable to use this token as they do not know its type.

So a default or any type would create "write only" tokens.
You could store them, but you would be unable to read them.
>

I agree with you. I'm always thinking as "when things go wrong". But in this case, ony if "type" becomes required, maybe it makes no sense to have also a defult type.

TravisSpomer:
I agree with this; I don't see the point to having `$type` be optional. It seems like it saves a couple of seconds of typing if you're manually creating a token file at the expense of making it harder for future humans *and* code to deal with.

c1rrus:
Perhaps the spec can be worded more clearly but: **Every token has an unambiguous type** and tools are expressly forbidden from trying to guess a type (e.g. from the value or name).

[...]

Issue #102, opened 2022-01-13
Typography type feedback


c1rrus:
Is the specification for typography styles that is being introduced by PR #86 fit for purpose? [Should the `lineHeight` sub-value use a number value, dimension or a new line-height type](https://github.com/design-tokens/community-group/pull/86#discussion_r768137006)?

mkeftz:
Thanks @c1rrus for the amazing work on composite types. It's been fantastic from a tool makers perspective to have these properly defined.

There are a couple of things on Typography I'm unsure of:
--  I feel it's missing a few frequently used properties. Specifically `fontStyle`, `textDecoration`, and `textTransform`. This makes it hard to translate between different design tools Text styles and design tokens. 
-- With the fontName property, is there a reason this is not fontFamily? This might be more of a personal preference, but all the other names align with their css equivalent.

ilikescience:
I think that the typography composite type serves as a good test balloon (or stress test, or canary in the coal mine, maybe) for some of the higher-level ideas about composite types.

For example, #86 defines a typography token as one that must include the following five values:
- fontName
- fontSize
- fontWeight
- letterSpacing
- lineHeight

However, in CSS alone, there are at least 57 properties that can change the way text is displayed.

<details><summary>CSS properties that modify text</summary>

- direction
- font
- font-family
- font-feature-settings
- font-kerning
- font-language-override
- font-optical-sizing
- font-palette
- font-size
- font-size-adjust
- font-stretch
- font-style
- font-synthesis
- font-synthesis-small-caps
- font-synthesis-style
- font-synthesis-weight
- font-variant
- font-variant-alternates
- font-variant-caps
- font-variant-east-asian
- font-variant-emoji
- font-variant-ligatures
- font-variant-numeric
- font-variant-position
- font-variation-settings
- font-weight
- glyph-orientation-vertical
- hanging-punctuation
- hyphens
- letter-spacing
- line-break
- overflow-wrap
- tab-size
- text-align
- text-align-all
- text-align-last
- text-combine-upright
- text-decoration
- text-decoration-color
- text-decoration-line
- text-decoration-style
- text-emphasis
- text-emphasis-color
- text-emphasis-position
- text-emphasis-style
- text-indent
- text-justify
- text-orientation
- text-shadow
- text-transform
- text-underline-position
- unicode-bidi
- white-space
- word-break
- word-spacing
- word-wrap
- writing-mode

</details>

[...]

Issue #199, opened 2022-12-12
Implicit vs. explicit typing in composite tokens


ilikescience:
The current format defines composite tokens in this way:

```json
{
  "shadow-token": {
    "$type": "shadow",
    "$value": {
      "color": "#00000088",
      "offsetX": "0.5rem",
      "offsetY": "0.5rem",
      "blur": "1.5rem",
      "spread": "0rem"
    }
  }
}
```

Unlike in a non-composite token, the values aren't explicitly typed. Take, for instance, in the composite token you have this:
```
...
    "offsetX": "0.5rem",
...
```

In another token, you'd write that as:

```json
{
    "horizontal offset": {
        "$type": "dimension",
        "$value": "0.5rem"
    }
}
```
Currently, the spec is written in such a way that the shadow composite token ALWAYS has a `offsetX` key, and that the `offsetX `key ALWAYS has a value of type `dimension`, so a parser should be able to correctly "guess" the type even though it's not explicitly stated.

However, I see a future where this fact gets hard to maintain. Take the typography composite token, for example.

In issue #102, folks have said that the current list of properties is too restrictive. At the same time, I pointed out that in CSS alone there are 57 possible properties that could apply to typography, and proposed a model of a required base set + optional extended set of properties.

In order for implicit typing to work, the has to list out a single unambiguous type for every property that the composite token might have. There's also the possibility that a property might have two allowed types, which causes some additional complexity.

The other option is to _explicitly_ type values of composite tokens:

```json
  "shadow-token": {
    "$type": "shadow",
    "$value": {
      "color": {
        "$type": "color",
        "$value": "#00000088"
      },
      "offsetX": {
        "$type": "dimension",
        "$value": "0.5rem"
      },
      "offsetY": {
        "$type": "dimension",
        "$value": "0.5rem"
      },
      "blur": {
        "$type": "dimension",
        "$value": "1.5rem"
      },
      "spread": {
        "$type": "dimension",
        "$value": "0rem"
      }
    }
  }
}
```

Obviously this is much harder to type. But the benefit is that the spec doesn't have to maintain the list of type mappings, the parser doesn't have to implement that map, and that composite tokens can have "open" definitions which have a small set of required properties but allow for any additional properties on top of that.

This also blurs the lines between a composite type and a group, which is a much bigger topic.

romainmenke:
This is very close to an alternative I proposed here : https://github.com/design-tokens/community-group/issues/126 👍

CITguy:
[...]

PR #201, opened 2023-01-03
Removes JSON types and makes absence of (inherited) $type invalid

### LINKED ISSUE #120: Remove JSON derived types from the specification
https://design-tokens.github.io/community-group/format/#types

This part of the specification assumes that implementers will have an inherent understanding of certain types and how to handle them.

This implicit typing will lead to interop issues.
It's better to have exact definitions for each type supported in the schema.

An obvious example is `String`.
When converting a value to CSS it is ambiguous if it should become an ident token or a string token.

```json
{
  "$value": "center"
}
```

```css
.foo {
  content: "center";
}
```

or

```css
.foo {
  text-align: center;
}
```

You also can't do this in the tokens file :

```json
{
  "$value": "\"center\""
}
```

Because that dictates how strings work in any consuming context.

c1rrus:
Closes #120, #139

* Removes support for the JSON types (`string`, `boolean`, `null`, etc.)
    * Except for `number` which is now documented as one of the spec's types. This is because it was required by the `gradient` type.
* Changes the type resolution rules so that when there is no (inherited) `$type` present, a token must be considered invalid. Previously, it would fall back on to one of the JSON types.
* Fixes a bug (?) in the `typography` type, where `lineHeight` expected a `string` value.
    * I had to change this since `string` no longer exists. But also, it was never a useful type of value for that purpose. I've changed it to expect `number` values for now. We can always revise it if needed once we review and action the feedback provided in #102 
* Updates examples throughout the spec to ensure that they are valid now that the JSON types and associated type resolution fallback behaviour have been removed.

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -416,7 +416,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
 - `fontSize`: The size of the typography. The value of this property MUST be a valid [dimension value](#dimension) or a reference to a dimension token.
 - `fontWeight`: The weight of the typography. The value of this property MUST be a valid [font weight](#font-weight) or a reference to a font weight token.
 - `letterSpacing`: The horizontal spacing between characters. The value of this property MUST be a valid [dimension value](#dimension) or a reference to a dimension token.
-- `lineHeight`: The vertical spacing between lines of typography. The value of this property MUST be a valid [dimension value](#dimension) or [number value](#number) (in which case it should be interpreted as a unitless line height), or a reference to a dimension or number token.
+- `lineHeight`: The vertical spacing between lines of typography. The value of this property MUST be a valid [number value](#number) or a reference to a number token. The number SHOULD be interpreted as a multiplier of the `fontSize`.
 
 <aside class="example" title="Typography composite token examples">
 
@@ -440,7 +440,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
         "fontSize": "{font.size.smallest}",
         "fontWeight": "{font.weight.normal}",
         "letterSpacing": "0px",
-        "lineHeight": "24px"
+        "lineHeight": 1
       }
     }
   }
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -374,6 +374,7 @@ Describes a gradient that is solid yellow for the first 2/3 and then fades to re
   },
 
   "position-end": {
+    "$type": "number",
     "$value": 1
   },
 
@@ -415,7 +416,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
 - `fontSize`: The size of the typography. The value of this property MUST be a valid [dimension value](#dimension) or a reference to a dimension token.
 - `fontWeight`: The weight of the typography. The value of this property MUST be a valid [font weight](#font-weight) or a reference to a font weight token.
 - `letterSpacing`: The horizontal spacing between characters. The value of this property MUST be a valid [dimension value](#dimension) or a reference to a dimension token.
-- `lineHeight`: The vertical spacing between lines of typography. The value of this property MUST be a valid JSON string or a reference to a string token.
+- `lineHeight`: The vertical spacing between lines of typography. The value of this property MUST be a valid [dimension value](#dimension) or [number value](#number) (in which case it should be interpreted as a unitless line height), or a reference to a dimension or number token.
 
 <aside class="example" title="Typography composite token examples">
 
@@ -429,7 +430,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
         "fontSize": "42px",
         "fontWeight": "700",
         "letterSpacing": "0.1px",
-        "lineHeight": "1.2"
+        "lineHeight": 1.2
       }
     },
     "microcopy": {
@@ -439,7 +440,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
         "fontSize": "{font.size.smallest}",
         "fontWeight": "{font.weight.normal}",
         "letterSpacing": "0px",
-        "lineHeight": "1"
+        "lineHeight": "24px"
       }
     }
   }
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -13,7 +13,7 @@
 }
 ```
 
-Note: The `$type` property has been added to ensure this example is valid. Please refer to the [type chapter](#type-0) for more details.
+Note: The `$type` property has been added to ensure this example is valid. Please refer to the [design token type chapter](#type-0) for more details.
 
 </aside>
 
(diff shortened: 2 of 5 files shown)


technical-reports/format/composite-types.md
```diff
@@ -439,7 +440,7 @@ Represents a typographic style. The `$type` property MUST be set to the string `
         "fontSize": "{font.size.smallest}",
         "fontWeight": "{font.weight.normal}",
         "letterSpacing": "0px",
-        "lineHeight": "1"
+        "lineHeight": "24px"
```
romainmenke, line 443: The type of `lineHeight` is ambiguous if both `1.2` and `"24px"` are allowed.

It is no longer possible to determine the value based on the current specification.
The specification only describes the `$type` property as a source for type information.

It is not possible to allow multiple types on the same field in a composite token.

romainmenke (changes requested):

technical-reports/format/composite-types.md
c1rrus: I think it's a bit of a grey area. The spec just says that "names and types of sub-values are pre-defined". That doesn't necessarily _forbid_ a composite type for specifying a range of permitted types for one of its sub-values.

However, I agree that this probably needs to be made more explicit one way or the other. I only updated the `typography` type in this PR because `string` is now gone. To avoid this PR getting side-tracked on updates to the `typography` type (which is not this PR's purpose), I can perhaps change it to only accept `dimension` values for now. That's hopefully a more "minimal" change that achieves the original aim of not replacing the `string` type with something that's still valid.

romainmenke: > I think it's a bit of a grey area. The spec just says that "names and types of sub-values are pre-defined". That doesn't necessarily forbid a composite type for specifying a range of permitted types for one of its sub-values.

This is definitely not a grey area :)
The specification doesn't define how a token parser must handle a composite token with fields that can have one of multiple types. Because it is not defined it can not be implemented. That it can not be implemented means that it isn't allowed/supported.

I think this issue is relevant : https://github.com/design-tokens/community-group/issues/199

-----

> I can perhaps change it to only accept dimension values for now.

Isn't `number` a better fit?
All the example previously were numbers but "embedded" in a JSON string.

I am unsure if this was something that leaked from CSS or not as it is typical for CSS that line height is a fraction of font size and therefor not a dimension.

c1rrus: I've changed it to just `number` for now. Like you said, it's closer to what we had previously.

romainmenke: Thank you for these changes 🙇

c1rrus: Thanks for your review comments :-)

technical-reports/format/design-token.md
```diff
@@ -7,19 +7,23 @@
...
 {
   "token name": {
-    "$value": "token value"
+    "$value": "#fff000",
+    "$type": "color"
   }
 }
 ```
 
+Note: The `$type` property has been added to ensure this example is valid. Please refer to the [type chapter](#type-0) for more details.
```
CITguy, line 16: Should the link be reworded to disambiguate between "5.4 Type", "6.1.2 Type", and "8 Types"?

e.g.,...

```markdown
... Please refer to the [design token type property chapter](#type-0) for more details.
```

technical-reports/format/design-token.md
```diff
@@ -113,7 +125,7 @@ The `$type` property can be set on different levels:
 - at the group level
 - at the token level
 
-The `$type` property MUST be a plain JSON string, whose value is `string`, `number`, `boolean`, `object`, `array`, `null` or one of the values specified in respective [type chapters](#types). The value of `$type` is case-sensitive.
+The `$type` property MUST be a plain JSON string, whose value is one of the values specified in respective [type chapters](#types). The value of `$type` is case-sensitive.
```
CITguy, line 128: * I'd recommend removing the link to `#types` ("8 Types"), purely because local context would require linking to BOTH "8 Types" and "9 Composite types".  
* Additionally, I'd reword it as `... specified in respective type definitions.`

With the above changes, if § 8 and 9 were to be merged/reorganized in the future, then this content would still be correct and we could potentially reintroduce the link later.

```markdown
...values specified in respective type definitions. The value ...
```

technical-reports/format/design-token.md
```diff
@@ -89,6 +100,7 @@ The value of the `$description` property MUST be a plain JSON string, for exampl
 {
```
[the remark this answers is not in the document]
c1rrus: It's the still correct that `$value` is the only required property on a design token object. `$type` can still be omitted, as long as one of the parent groups has `$type`, or it's a reference to another token.

technical-reports/format/design-token.md
c1rrus: Agreed. I'll make that change in a moment.

technical-reports/format/design-token.md
c1rrus: Good spot! I'll change it along the lines you've suggested.

Longer term, I think quite a few spec chapters could probably do with some re-organising. But that's for another day... ;-)

technical-reports/format/design-token.md
```diff
@@ -113,7 +125,7 @@ The `$type` property can be set on different levels:
 - at the group level
 - at the token level
 
-The `$type` property MUST be a plain JSON string, whose value is `string`, `number`, `boolean`, `object`, `array`, `null` or one of the values specified in respective [type chapters](#types). The value of `$type` is case-sensitive.
+The `$type` property MUST be a plain JSON string, whose value is one of the values specified in this specification's respective type definitios. The value of `$type` is case-sensitive.
```
romainmenke, line 128:
```suggestion
The `$type` property MUST be a plain JSON string, whose value is one of the values specified in this specification's respective type definitions. The value of `$type` is case-sensitive.
```

typo :)

technical-reports/format/design-token.md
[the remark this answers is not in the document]
kevinmpowell: @c1rrus I think it would be clearer to state that type is required along with name and value. Something like:

> Name, value and type are required. `$type` can be omitted from an individual token definition if defined on a parent group. See more information on Types below.

dbanksdesign (approved):
LGTM! Thank you for putting this together.

ilikescience (approved):
LGTM!

technical-reports/format/design-token.md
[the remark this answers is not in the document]
c1rrus: @kevinmpowell If it's ok with you, I'd like to leave this as is and merge this PR given that @dbanksdesign & @ilikescience both approved it now.

I think this is a symptom of a broader issue that would be best addressed in a separate PR: The spec's current use of the word "property" is overloaded. There are:

- _Syntax_ properties - i.e. the `$whatever` properties that are written in the files.
- _Model_ properties (someone please suggest a better name! 😛) - the properties that the (parsed) tokens and groups have.

Perhaps we need to agree on terminology to differentiate between these two meanings and then apply that consistently throughout the whole spec.

I'd argue that the for the current statement that "$value is the only required property for a token" is correct in the _syntax_ sense. However, in the _model_ sense, every token has at least name, value and type.

PR #202, opened 2023-01-04
Adds extensions to groups

### LINKED ISSUE #106: Can extensions be applied on the group level?
I've been building [Token CSS](https://github.com/tokencss/tokencss), a CSS tool based on the current draft spec.

I really like the groups can define a `type` which will be applied to all child tokens. I was curious if `extensions` could also be applied to groups? I'm hoping to use `extensions` to define a `scale` ([`semantic-type`](https://github.com/design-tokens/community-group/issues/104) or `used-for` values) on a per-group basis.

I'm not sure there's anything blocking this usage in the current specification but clarity would be nice! I don't think the spec needs to weigh in on how inheritance/cascading of `extensions` could work—that seems like an implementation detail left up to each tool.

Edit: I had not seen that https://github.com/design-tokens/community-group/pull/89 moves group-level properties under the `metadata` name, but I'm a huge fan of that move! I guess my question then becomes "Is `extensions` allowed inside of `group.metadata`?"

c1rrus:
Closes #106 

There seems to be a consensus that allowing groups to have `$extensions` (just like design tokens can), so this updates the spec accordingly.

### DIFF
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -167,7 +167,7 @@ Group extensions follow the same rules as [design token extensions](#extensions)
 - The keys SHOULD be chosen such that they avoid the likelihood of a naming clash with another vendor's data. The [reverse domain name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) is recommended for this purpose.
 - Tools that process design token files MUST preserve any extension data they do not themselves understand. For example, if a group contains extension data from tool A and the file containing that data is opened by tool B, then tool B MUST include the original tool A extension data whenever it saves a new design token file containing that group.
 
-Note that, since a group's `$extensions` only relate to that group, they do not have any effect on nested groups or tokens. However, if an extension author wants their extension to be inherited by nested groups or tokens like a group's `$type` is, then they MAY implement that behaviour in whatever tools support their extension.
+Note that, since a group's `$extensions` only relate to that group, they do not have any effect on nested groups or tokens.
 
 <aside class="example">
 
--- a/technical-reports/format/groups.md
+++ b/technical-reports/format/groups.md
@@ -130,12 +130,6 @@ Suggested ways tools MAY use this property are:
 - A GUI tool that lets users browse or select tokens could display this info alongside the corresponding group or as a tooltip
 - Translation tools could output this as a source code comment
 
-<div class="issue" data-number="72">
-
-Groups may support additional properties like type and description. Should other properties be supported at the group level?
-
-</div>
-
 ### Type
 
 Groups MAY include an optional `$type` property so a type property does not need to be manually added to every token. [See supported "Types"](#types) for more information.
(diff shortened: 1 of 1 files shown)


technical-reports/format/groups.md
```diff
@@ -164,6 +158,46 @@ For example:
...
+
+Group extensions follow the same rules as [design token extensions](#extensions), the only difference is that they relate to the group itself.
+
+- The keys SHOULD be chosen such that they avoid the likelihood of a naming clash with another vendor's data. The [reverse domain name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) is recommended for this purpose.
+- Tools that process design token files MUST preserve any extension data they do not themselves understand. For example, if a group contains extension data from tool A and the file containing that data is opened by tool B, then tool B MUST include the original tool A extension data whenever it saves a new design token file containing that group.
+
+Note that, since a group's `$extensions` only relate to that group, they do not have any effect on nested groups or tokens. However, if an extension author wants their extension to be inherited by nested groups or tokens like a group's `$type` is, then they MAY implement that behaviour in whatever tools support their extension.
```
kevinmpowell, line 170: @c1rrus I'm concerned about the ambiguity here. If one vendor's `$extensions` are inherited by children but another vendor's aren't, I'm concerned about our ability to "pave cowpaths" in the future.

My expectation would be that group `$extensions` are *not* inherited by child groups or tokens.

kevinmpowell: This also seems to be the general consensus in the original issue: #106.

I'd delete the sentence that starts with "However, if an extension author..." and call this one done. 

PR looks good otherwise.

lauthieb (approved):
Completely agree 💯
Thanks @c1rrus

technical-reports/format/groups.md
```diff
@@ -164,6 +158,46 @@ For example:
...
+### Extensions
+
+Groups MAY include an optional `$extensions` property where tools MAY add proprietary, user-, team- or vendor-specific data to a group. When doing so, each tool MUST use a vendor-specific key whose value MAY be any valid JSON data.
+
+Group extensions follow the same rules as [design token extensions](#extensions), the only difference is that they relate to the group itself.
+
+- The keys SHOULD be chosen such that they avoid the likelihood of a naming clash with another vendor's data. The [reverse domain name notation](https://en.wikipedia.org/wiki/Reverse_domain_name_notation) is recommended for this purpose.
+- Tools that process design token files MUST preserve any extension data they do not themselves understand. For example, if a group contains extension data from tool A and the file containing that data is opened by tool B, then tool B MUST include the original tool A extension data whenever it saves a new design token file containing that group.
+
+Note that, since a group's `$extensions` only relate to that group, they do not have any effect on nested groups or tokens.
```
pascalduez, line 170: I'm wondering whether this sentence is straightforward enough. 

Some parts of #106 feels a bit more explicit (too much?)
> Groups `$extensions` should not be inherited. i.e. tokens and groups within a group that has `$extensions` should not be treated as though they themselves had the same $extensions property.

PR #207, opened 2023-03-09
Update fontWeight to be number in Example 36

origami-z:
According to [font weight](https://design-tokens.github.io/community-group/format/#font-weight) section, it should be number instead of string `"700"`


c1rrus (approved):
Good spot! Thanks for contributing this correction. 🙌

PR #242, opened 2024-08-16
Chore: Upgrade Node to v22, update deps

drwpow:
## Changes

- Updates Node from `16` → `22`. Node 16 is EOL. Welcome to the future 😎 
- Updates deps. Because there were install errors for old packages on modern Node versions, and also because there were multiple security warnings.
- Updates GitHub Actions to current versions.

## Reviewing

- Look at [Preview Link](https://deploy-preview-242--dtcg-tr.netlify.app/)
- Look at GitHub logs

PR #243, opened 2024-08-16
Update GitHub Actions

drwpow:
⚠️ Depends on #242 

## Changes

Updates GitHub Action runners to the current version. No other changes.

## Reviewing

- GitHub Actions will warn us of an error… right?


.github/workflows/pr.yml
```diff
@@ -11,16 +11,17 @@ env:
...
-      - uses: actions/cache@v2
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v4
         with:
           path: ~/.npm
           key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-node-
-      - uses: actions/setup-node@v3
-        with: { node-version: 16, cache: npm }
```
drwpow, line 24: Note: removing `node-version` outright should be fine; we just get the latest, when it updates. Back in the days of 10.x/12.x/14.x/16.x, changes were sweeping, and significant. Nowadays the major versions probably won’t break anything (this repo relies on fairly-minimal stuff)

drwpow: CleanShot 2024-08-16 at 15 51 29@2x

Whoa using a newer version of Node in GitHub Actions (and newer actions) sped it up by quite a bit! The newer version completed before the old version did (and the old version got a head start, too!)

drwpow:
On second thought, closing. #242, when we update versions, fails in GitHub Actions. Better to do both at once.

Issue #218, opened 2023-05-06
Remove REM/EM from specification?


caoimghgin:
I've noticed font size values are spec'd as REM for W3C. Design token values should be platform agnostic. Native apps such as Android and iOS have no such concept, and the declaration of REM must point to another base value for rem to be meaningful.

nateclicks:
Is there a different solution you have in mind to replace it with or a different relative unit that is platform-agnostic?

They have provided a unit for a absolute values (px) and a unit for a relative values (rem) in the spec and have noted that a translation tool will need to provide appropriate conversions when preparing tokens for a specific platform.

There is definitely value to having absolute and relative values available for dimensions so if there is another direction that is better, it'd be great look into it.

caoimghgin:
IF they provided absolute units in specs, a  fontSize: '16px' output, I'm all good. Please close request.

StfBauer:
REM and EM are appropriate values, while the pixel is not:

Apple list its default font sizes on iOS.
https://developer.apple.com/design/human-interface-guidelines/typography

Body text, on default, refers to 17pt, which is one rem; depending on the user's preferences, the value for one rem or one em changes for accessibility.

Android:
Material design has a conversion table in their guidelines on moving from REM to platform-specific values.
https://m3.material.io/styles/typography/type-scale-tokens

To use pixel values in the case of design tokens is the wrong approach these days. REM and EM values, on the other hand, are quasi-the-platform-agnostic units that most design systems also relate to.

caoimghgin:
A traditional px (or point) is the appropriate value, where REM/EM are derived values that are driven by the idea of a CSS Pixel. 

I believed 1pt was always equal to 1px (1/72") because iOS specifies typography in points, and that is how print works. I am confirmed in my belief when I set spec typography as size 17 (in Sketch, Figma, or Illustrator) as a developer handoff, I can screenshot the render of that design on my iOS device and expect it to overlay perfectly on my designs. Even though Illustrator says the typography is in points and documentation in Figma says it's in px, both line up perfectly.

The problem occurs only on the web, with the idea of a CSS Pixel. This strange concept gives us math such as... 

`17pt == 22.66px == 1.4167rem `

This ASSUMES a root font-size of 16px, and as we know, that is a variable that could be changed to anything in the web dev environment.

Therefore, we should consider using the unit specification that matches the input defined by the designer in a design application as 'the standard', and leave the job to programmers to transform to any platform-specific values they like via Style Dictionary transforms..

[...]

Issue #121, opened 2022-03-25
Remove units from type definitions


romainmenke:
https://design-tokens.github.io/community-group/format/#duration

This assumes that all consumers agree on the unit for a given value type or forces implementers to do error prone parsing before passing on transformed values.

A better sub type might be a general `Dimension` type

```json
{
  "Duration-100": {
    "$value": {
        "number": "100",
        "unit": "ms"
    },
    "$type": "duration"
  }
}
```

This reduces the amount of micro syntaxes in the spec and makes it easier to implement.

drwpow:
Parsing numbers and units from a string is trivial and performant in every programming language. I personally quite like the readability of having one string represent amount + unit, and it doesn’t impact tooling. So I think this hurts readability without adding any benefit (I don’t think this is easier to implement).

Further, for duration, there’s only one unit (considering `ms` and `s` are just divisions of the same unit), so expanding it just seems to be verbose.

> A better sub type might be a general `Dimension` type

In a design system, time and space are distinct concepts. Numbers can quantify both, but since they have no direct relationship I don’t think there’s a point in tokenizing the number (e.g. `50ms` and `50px` have no relationship).

TravisSpomer:
Agreed. I think that what the spec defines is our best option right now, unless *maybe* perhaps we could agree on something like "all durations shall be in milliseconds forever so let's specify durations as just a `number`." But since the spec already allows both `px` and `rem` for dimensions, I think that durations should probably be strings that specify units too.

jonathantneal:
I want to drop a little “it depends” to this conversation, but more for consideration in future issues.

Given the current [character restrictions](https://design-tokens.github.io/community-group/format/#character-restrictions), I agree we should be able to reliably parse units from numbers in the JSON format.

> Parsing numbers and units from a string is trivial and performant in every programming language.

I only want to highlight one exception, specific to this Design Tokens Format Module. These character restrictions do not pair well outside of quoted strings in non JSON file formats like CSS.

Specifically, the reservation of `U+002E` — the **FULL STOP** or ‘period’ (`.`) character — may potentially introduce non-trivial parsing issues. This is because the period character also begins fractional numbers.

[...]

Issue #200, opened 2022-12-12
Define error handling


romainmenke:
see : https://github.com/design-tokens/community-group/issues/121#issuecomment-1346869968

Defining which error cases can be expected while parsing a tokens file and what to do for each error is critical.

Errors carve out space for future extensions.

A simplistic example :

`foo` has a declared type of `string` but a numeric value.

```json
{
  "foo": {
    "$type": "string"
    "$value": 10
  }
}
```

- this MUST be an error.
- implementations MUST NOT cast to a string (`"10"`)
- translation tools MUST NOT generate code based on this token
- design tools MUST NOT expose this token to designers

If we allow tools to be forgiving and help authors by casting types or passing on invalid tokens we reduce the available space for future extension.

https://tr.designtokens.org/format/#typography

What if we want to add number values for `lineHeight` but give them specific meaning?
This is only possible if number values are not are in use by accident.

-----

I don't have a strong opinion about the best definition of error handling for design tokens.
But I think it is important to have some definition and then test each token type against this definition.

The main questions are :
- what if the type and value do not match?
- what if a sub value in a micro syntax is invalid/unknown?
- what if a value does not match the known options from an enum?
- what if a composite token has an unexpected shape?
- how does an invalid result affect the outcome?
- does it require a breaking change in the specification to support emerging technologies?

These questions must be seen from the perspective of a mature ecosystem.
- we have reached specification version X
- there are hundreds of tools, some only support an older version of the specification
- organizations use many different tools together

davejsdev:
The most recent draft of the design token format now has guidance for error conditions: https://www.designtokens.org/tr/drafts/format/

@romainmenke does the current spec draft fully address your issue?

PR #244, opened 2024-08-16
Proposal: changes to type: $dimension and type: $duration tokens

drwpow:
## Summary

This proposes the following changes to `$type: dimension` and `$type: duration` tokens, which have the same designs:

- Changes `$value` from a string → object with `value` and `unit` keys for both dimension and duration

## Reasoning

This PR proposes that the following discussions be resolved like so:

- #121: ✅ **Accepted** (as written)
- #218: ❌ **Rejected** (that moves in the opposite direction, turning dimension/durations into number-only values)

## Pros

- Simplifies tooling because these token types would no longer need string parsing (which could potentially introduce bugs)
  - Other types like shadows and borders already split their values into an object with keys
- Reduces errors in output.
  - For example, if someone was exporting dimension tokens to `.js`, they could normalize everything to `px` and just output numbers (which is the default behavior of JS styling). Tooling wouldn’t have to try and parse the string from the number, potentially introducing bugs.
- Simplifies validation because string pattern matching is no longer required
  - Validation is now as simple as checking for the object wrapper, both keys, and both keys match their expected primitive types.

## Cons

- Objects are more verbose to type

## Alternatives 

- We could be more lax about `$value.unit` for dimensions, however, that seems to introduce more problems and regress that part of the schema ([see comment below](https://github.com/design-tokens/community-group/pull/244#issuecomment-2294724417))
- We could accept #218, and downconvert all dimensions into `px` values and durations into `ms`. But this would break many setups in tokens (chiefly, web—`rem`s and `px` are distinct values with different uses, and based on implementation, aren’t always mathematically-equivalent).
  - We could also split the difference, and keep the object for dimensions, and just number for duration (standardizing on milliseconds only). That’s just “syntactic sugar” anyway.

## Notes

- The original proposal had `$value.number` instead of `$value.value`, possibly to reduce duplication but I didn’t find a rationale. This changes to `value` because that’s a more common nomenclature for this term, e.g. `csstree`: 

  screenshot of csstree showing a “value” / “unit” object

(PR description edited from original version to reflect changes in code)

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -19,10 +19,10 @@ A design token whose type happens to be a composite type is sometimes also calle
     "$type": "shadow",
     "$value": {
       "color": "#00000080",
-      "offsetX": "0.5rem",
-      "offsetY": "0.5rem",
-      "blur": "1.5rem",
-      "spread": "0rem"
+      "offsetX": { "value": 0.5, "unit": "rem" },
+      "offsetY": { "value": 0.5, "unit": "rem" },
+      "blur": { "value": 1.5, "unit": "rem" },
+      "spread": { "value": 0, "unit": "rem" }
     }
   }
 }
@@ -37,7 +37,7 @@ A design token whose type happens to be a composite type is sometimes also calle
   "space": {
     "small": {
       "$type": "dimension",
-      "$value": "0.5rem"
+      "$value": { "value": 0.5, "unit": "rem" }
     }
   },
 
@@ -56,8 +56,8 @@ A design token whose type happens to be a composite type is sometimes also calle
         "color": "{color.shadow-050}",
         "offsetX": "{space.small}",
         "offsetY": "{space.small}",
-        "blur": "1.5rem",
-        "spread": "0rem"
+        "blur": { "value": 1.5, "unit": "rem" },
+        "spread": { "value": 0, "unit": "rem" }
       }
     }
   },
@@ -141,7 +141,10 @@ Object stroke style values MUST have the following properties:
   "alert-border-style": {
     "$type": "strokeStyle",
     "$value": {
-      "dashArray": ["0.5rem", "0.25rem"],
+      "dashArray": [
+        { "value": 0.5, "unit": "rem" },
+        { "value": 0.25, "unit": "rem" }
+      ],
       "lineCap": "round"
     }
   }
@@ -157,14 +160,17 @@ Object stroke style values MUST have the following properties:
   "notification-border-style": {
     "$type": "strokeStyle",
     "$value": {
-      "dashArray": ["{dash-length-medium}", "0.25rem"],
+      "dashArray": ["{dash-length-medium}", { "value": 0.25, "unit": "rem" }],
       "lineCap": "butt"
     }
   },
 
   "dash-length-medium": {
     "$type": "dimension",
-    "$value": "10px"
+    "$value": {
+      "value": 10,
+      "unit": "px"
+    }
   }
 }
 ```
@@ -214,7 +220,10 @@ Represents a border style. The `$type` property MUST be set to the string `borde
       "$type": "border",
       "$value": {
         "color": "#36363600",
-        "width": "3px",
+        "width": {
+          "value": 3,
+          "unit": "px"
+        },
         "style": "solid"
       }
     },
@@ -222,9 +231,12 @@ Represents a border style. The `$type` property MUST be set to the string `borde
       "$type": "border",
       "$value": {
         "color": "{color.focusring}",
-        "width": "1px",
+        "width": {
+          "value": 1,
+          "unit": "px"
+        },
         "style": {
-          "dashArray": ["0.5rem", "0.25rem"],
+          "dashArray": [{ "value": 0.5, "unit": "rem" }, "0.25rem"],
           "lineCap": "round"
         }
       }
@@ -255,8 +267,8 @@ Represents a animated transition between two states. The `$type` property MUST b
     "emphasis": {
       "$type": "transition",
       "$value": {
-        "duration": "200ms",
-        "delay": "0ms",
+        "duration": { "value": 200, "unit": "ms" },
+        "delay": { "value": 0 },
         "timingFunction": [0.5, 0, 1, 1]
       }
     }
@@ -289,10 +301,10 @@ Represents a shadow style. The `$type` property MUST be set to the string `shado
     "$type": "shadow",
     "$value": {
       "color": "#00000080",
-      "offsetX": "0.5rem",
-      "offsetY": "0.5rem",
-      "blur": "1.5rem",
-      "spread": "0rem"
+      "offsetX": { "value": 0.5, "unit": "rem" },
+      "offsetY": { "value": 0.5, "unit": "rem" },
+      "blur": { "value": 1.5, "unit": "rem" },
+      "spread": { "value": 0, "unit": "rem" }
     }
   },
   "layered-shadow": {
@@ -336,6 +348,10 @@ Represents a shadow style. The `$type` property MUST be set to the string `shado
 ```
 
 </aside>
+
+<div class="issue" data-number="100" title="Shadow type feedback">
+  Is the current specification for shadows fit for purpose? Does it need to support multiple shadows, as some tools and platforms do?
+</div>
 
 ## Gradient
 
@@ -461,9 +477,15 @@ Represents a typographic style. The `$type` property MUST be set to the string `
       "$type": "typography",
       "$value": {
         "fontFamily": "Roboto",
-        "fontSize": "42px",
+        "fontSize": {
+          "value": 42,
+          "unit": "px"
+        },
         "fontWeight": 700,
-        "letterSpacing": "0.1px",
+        "letterSpacing": {
+          "value": 0.1,
+          "unit": "px"
+        },
         "lineHeight": 1.2
       }
     },
@@ -473,7 +495,10 @@ Represents a typographic style. The `$type` property MUST be set to the string `
         "fontFamily": "{font.serif}",
         "fontSize": "{font.size.smallest}",
         "fontWeight": "{font.weight.normal}",
-        "letterSpacing": "0px",
+        "letterSpacing": {
+          "value": 0,
+          "unit": "px"
+        },
         "lineHeight": 1
       }
     }
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -236,7 +236,10 @@ Represents a border style. The `$type` property MUST be set to the string `borde
           "unit": "px"
         },
         "style": {
-          "dashArray": [{ "value": 0.5, "unit": "rem" }, "0.25rem"],
+          "dashArray": [
+            { "value": 0.5, "unit": "rem" },
+            { "value": 0.25, "unit": "rem" }
+          ],
           "lineCap": "round"
         }
       }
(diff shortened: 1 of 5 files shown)


technical-reports/format/types.md
```diff
@@ -60,22 +65,29 @@ For example:
...
+    },
     "$type": "dimension"
   }
 }
 ```
 
 </aside>
 
-The "px" and "rem" units are to be interpreted the same way they are in CSS:
+### Validation
```
drwpow, line 86: Note: this new subheading is **inconsistent** with the other types. This was done thinking back to #200 that observes the spec could be more explicit about validation.

This would be a change from the current structure of validation rules existing in prose, usually in the initial description. However, in many places, those aren’t exhaustive/complete (for good reason—it’d be hard to read).

All that said, we don’t have to introduce these subheadings in this PR; we could fold everything into prose form to match the existing format.

dev-nicolaos:
I like this proposal, it seems likely to make parsing easier for tools without introducing much burden on those who'd prefer to write tokens by hand. 

Making dimension units unrestricted seems like a win for platform agnosticism too. The web has so many dimension units and this lets token maintainers take advantage of that without introducing a bunch of web specific stuff to the spec.

I wonder if the `value` property names couldn't use a little bikeshedding but I can't think of anything better right now.

romainmenke:
Thank you for this proposal @drwpow!
Overall this looks good to me.

------

> Dimensions: no restrictions on what $value.unit can be (author-defined)

We can't have unrestricted units.

We will end up having these issues:
- tools will have conflicting unit definitions, fracturing the ecosystem
- tools will squat units that should be in the spec itself, but with a different definition than desired for the specification

Instead we should have a fixed set of specified units and allow any other unit as long as it is vendor prefixed.

Then there can't be conflicts between tools as those would be different vendors.
There also can't be conflicts between tools and future spec expansion.

It still leaves plenty of room for custom definitions.

Keep in mind that design token files are a carrier format.
How a unit is encoded in a file doesn't dictate how it is exposed in a design tool or the final output for developers.

Figma can create a custom unit `y` that is exposed to designers as `y` and output to developers as `y` while encoding it as `figma-y` in token files.

If a custom, vendor prefixed, unit is popular and gains adoption among other tools it is also a clear signal that it is useful. Such a custom unit should then go through the spec process to standardize it.

drwpow:
@romainmenke Those are all great points, thanks for raising. I was originally trying to solve for multiple things here, but agree that may be too big a change here:

- Tools may interpret units differently anyway (e.g. `10px == 1rem`). But to your point, the specification should still define these values because then it’s clear the tool doesn’t match the spec
- I was scared the review process of adding more units may take too long, and many other units hadn’t been been proposed.

But you address those nicely with the `vendor` addition proposal. I really like this version specifically for the following reasons:

```json
{
  "value": 10,
  "unit": "y",
  "vendor": "figma"
}
```

- I agree not using `$extensions` here because that is not part of the value, and shouldn’t affect interpretation of the value in every tool
- Having some sort of prefix in `unit` is dicey because it brings back string parsing/pattern matching that this PR tries to move away from that puts it back on tooling. Defining what a valid or invalid prefix is seems like a can of worms.
- The inclusion of `vendor` as an optional property is great because it “disables” unit validation in a sense, but without disabling it in every scenario.
- We also have other tokens (including @resource11’s upcoming proposal to the color format) where the addition of one object key changes the interpretation of the other values (it’s all one “value”)

Minor point: I would assume that if the following were provided:

```json
{
  "value": 10,
  "unit": "px",
  "vendor": "foobar"
}
```

This would NOT be interpreted as `px` as-defined by the spec. Because if, say, the intent was that it’s parsed as a `px` dimension by tooling by default, but a particular tool wants to convert it differently, then that’s the case where `$extensions` should be used instead (all tooling operates in a standard way, but in one particular tool, additional metadata overrides behavior). So am I correct in understanding `unit` + `vendor` = a unique unit that has no relation to any other?

drwpow:
But all that said, I’ll make this PR easier to review by **NOT** adding vendor here, and we could review that in a followup. I’ll restore the current behavior of only allowing `px` and `rem` and keeping those as currently-defined. Will also update the PR + description.

🤔 I think this still rejects #218 because that proposes numbers-only values.

romainmenke:
>  So am I correct in understanding unit + vendor = a unique unit that has no relation to any other?

Yes :)

technical-reports/format/types.md
```diff
@@ -157,18 +175,26 @@ For example:
...
-    "$value": "100ms",
+    "$value": {
+      "value": 100,
+      "unit": "ms"
+    },
     "$type": "duration"
   },
   "Duration-200": {
-    "$value": "200ms",
+    "$value": { "value": 200 },
```
c1rrus, line 185: If we make unit required, this would need updating.

c1rrus (commented):
Thanks for opening this PR @drwpow. It looks good. I spotted a few minor inconsistencies though - see the other review comments.

However, I'm on the fence about whether or not dimension and duration values need to be changed to an object format. Especially since it's a breaking change (i.e. files and tools written the current draft spec will need to be updated if this goes in). Not a deal breaker since we're still in "editor's draft" territory, but also not something to take lightly given that there are a number of DTCG tools and files out there in the wild.

Making parsing easier is nice, but I'm not convinced the current string values are a significant blocker or source of bugs for implementors. The more significant improvement if this were to go in is more consistency within the DTCG format's syntax. Especially since color values are also due to change to an object structure in order to support more color spaces and wider gamuts. I think aligning all type's value syntaxes around some JSON-esque rather than having some encode multiple pieces of info into a string and others not.

As you've noted this does make things more verbose for humans wanting to read or hand-edit DTCG files. However, I think the proposed syntax is still understandable. Also, I guess this would make creating a JSON schema for the DTCG format easier, and thus also make it easier auto-complete suggestions in editors like VSCode. So on balance, I think it could be a worthwhile trade off.

Nonetheless, I'd like some other format editors to take a look and chime in with their thoughts before I approve this PR.

c1rrus:
Btw, i agree with the earlier point from @romainmenke about keeping a defined set of allowed units. Without that I fear we'd quickly end up in a world of incompatible tools and files due different folks supporting different units, or interpreting the same unit in different ways.

I am intrigued and potentially in favour of adding support for custom units though. But, agree that should be done separate from this PR.

technical-reports/format/composite-types.md
```diff
@@ -255,8 +267,8 @@ Represents a animated transition between two states. The `$type` property MUST b
     "emphasis": {
       "$type": "transition",
       "$value": {
-        "duration": "200ms",
-        "delay": "0ms",
+        "duration": { "value": 200, "unit": "ms" },
+        "delay": { "value": 0 },
```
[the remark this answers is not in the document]
drwpow: Oh oh sorry, I got my comments mixed up. You’re right. The first cut of this played around with making `"unit": "ms"` optional, but the more I thought about it, that didn’t make sense. I don’t personally think it added clarity or improved UX, so I reverted it. But I inconsistently reverted it, as your comments caught. Thank you!

technical-reports/format/types.md
```diff
@@ -148,7 +161,12 @@ Example:
...
 
-Represents the length of time in milliseconds an animation or animation cycle takes to complete, such as 200 milliseconds. The `$type` property MUST be set to the string `duration`. The value MUST be a string containing a number (either integer or floating-point) followed by an "ms" unit. A millisecond is a unit of time equal to one thousandth of a second.
+Represents the length of time in milliseconds an animation or animation cycle takes to complete, such as 200 milliseconds. The `$type` property MUST be set to the string `duration`. The value MUST be an object containing a numeric `value` (either integer or floating-point) and a `unit` of milliseconds (`"ms"`) or seconds (`"s"`). A millisecond is a unit of time equal to one thousandth of a second.
+
+| Key     |   Type   | Required | Description                                                          |
+| :------ | :------: | :------: | :------------------------------------------------------------------- |
+| `value` | `number` |    Y     | An integer or floating-point value representing the numeric value.   |
+| `unit`  | `string` |    Y     | Unit of time. Supported values: `"ms"` (millisecond), `"s"`(second). |
```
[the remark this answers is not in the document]
drwpow: Yes! I agree. I had just been trying it on for size where `unit` was optional for duration (maybe as an imagined olive branch to people who wouldn’t like the syntax change). But after thinking about it more I agree just always making `unit` required is clear, and doesn’t ask too much of everyone.

Please keep finding inconsistencies! It is all helpful!

drwpow:
> Making parsing easier is nice, but I'm not convinced the current string values are a significant blocker or source of bugs for implementors.

So actually re-reading that thread, #121, there wasn’t a specific comment that touched on this, but comments like [@jonathantneal](https://github.com/design-tokens/community-group/issues/121#issuecomment-1085228851)’s reminded me that parsing can be trickier than it seems.

In JavaScript, validating that token does require a RegEx, which isn’t too bad. But when you need to separate it into number and unit, it is trickier than it seems at first glance. In JS, for one, you have `parseFloat()` and `parseInt()`, and one may parse as an integer when it’s not. But then you also have `BigInt`!

For other languages (that aren’t JS) it gets harder. For example, in Rust, [Serde](https://serde.rs/) is a popular library for interpreting JSON into Rust data structures and can automate a lot of the mapping and memory management. It gets trickier when you’re storing values that are delivered as strings, but they ultimately **want to be numbers with metadata.**

And of course that’s a shared problem among all languages: **storing those numbers and units somewhere.** After parsing, now your tooling has to have additional scaffolding to store those saved numbers and units separately from the original token JSON, and keep the mapping between the two preserved.

I think in hindsight, offloading all that work onto tooling doesn’t make sense, when even CSS ASTs like `csstree` separate values & units so they can be validated independently. It works when we’re plugging tokens straight into CSS, but for any other platform and language, it requires a lot more work than I think we originally realized.

kevinmpowell:
Any ideas on how this would support [unitless line-height](https://css-tricks.com/almanac/properties/l/line-height/#aa-unitless-line-heights)? Would the unit just be an empty string in that case?

romainmenke:
> Any ideas on how this would support [unitless line-height](https://css-tricks.com/almanac/properties/l/line-height/#aa-unitless-line-heights)? Would the unit just be an empty string in that case?

That is just a `number`, right?
Not a dimension.

I don't think it is needed to add dimensions with a "null unit" as a special form of number when regular numbers exist :)

ChucKN0risK:
Thanks for creating this PR @drwpow 👍

I think splitting the value to an object format is the best solution. I still easy to read while it makes it easier to parse for other languages and tools [as you explained](https://github.com/design-tokens/community-group/pull/244#issuecomment-2298998450).

I also think we should create a new PR to introduce the `vendor` property.

And finally I agree with @romainmenke about the fact we don't need to add dimensions with a "null unit". Unitless line height can be created as `number` tokens.

technical-reports/format/types.md
```diff
@@ -60,22 +65,30 @@ For example:
...
-- **rem**: Represents a multiple of the system's default font size (which MAY be configurable by the user). 1rem is 100% of the default font size. The equivalent of 1rem on Android is 16sp. Not all platforms have an equivalent to rem, so translation tools MAY need to do a lossy conversion to a fixed px size by assuming a default font size (usually 16px) for such platforms.
+- `$value.unit` may only be `"px"` or `"rem"`.
+  - **px**: Represents an idealized pixel on the viewport. The equivalent in Android is `dp` and iOS is `pt`. Translation tools SHOULD therefore convert to these or other equivalent units as needed.
+  - **rem**: Represents a multiple of the system's default font size (which MAY be configurable by the user). 1rem is 100% of the default font size. The equivalent of 1rem on Android is 16sp. Not all platforms have an equivalent to rem, so translation tools MAY need to do a lossy conversion to a fixed px size by assuming a default font size (usually 16px) for such platforms.
+- `$value.unit` is still requird even if `$value.value` is `0`.
```
SpikyC, line 91:
```suggestion
- `$value.unit` is still required even if `$value.value` is `0`.
```

drwpow:
DTCG meeting just met and this proposal has been approved! 🎉. Will update this PR with all the suggested changes

Issue #100, opened 2022-01-13
Shadow type feedback


c1rrus:
Is the [shadow composite type](https://design-tokens.github.io/community-group/format/#shadow) fit for purpose? Does it need to support multiple shadows, as some tools and platforms do?

Please share your feedback, thoughts and ideas in this issue.

drwpow:
I like it. That makes logical sense to me. My only request, personally, is that arrays of shadow objects are accepted [example](https://tobiasahlin.com/blog/layered-smooth-box-shadows/). IMO layering shadows still counts as “one shadow” and they are difficult/unwieldy to combine as separate tokens (i.e. they aren’t reusable because they depend on the precise position of other shadows in the array)

natemoo-re:
Totally agree with @drwpow that supporting arrays of shadow objects is critical for my use cases.

Is the idea that `color` tokens also encode an `alpha` value? Personally I would love to see an optional `opacity` or `alpha` value on the `shadow` object so that color tokens wouldn't need to be created for each opacity.

WardMe:
I don't see an `inset` property on shadows? How would we define box vs inner shadows?

TravisSpomer:
Yes, that appears to be an oversight. Seems like an `inset` property would be required.

Chudesnov:
> I would love to see an optional opacity or alpha value on the shadow object so that color tokens wouldn't need to be created for each opacity.

@natemoo-re maybe I'm overgeneralising, but to me it seems that a color part of a shadow is still a color token (even though it's "anonymous" and probably should never be referenced by anyone from the outside), it seems less then ideal to have separate definition mechanisms for the same types of tokens in different contexts.

Maybe, instead of that, sub-properties for composite tokens' `$value` should allow plain values as a shorthand but generally expect an anonymous token definition (`{ $value, etc... }`); then opacity/alpha/mixing transformation could be defined locally under that anonymous token.

Here's an example of what I mean (based on a different but similar spec we've been working on for some time at our company):

```
{
  "color": {
    "accent": {
      "$type": "color",
      "$value": "#000000"
    }
  },
  "shadow-token": {
    // most Web-compatible implementation would need support for a multi-layered shadow,
    // so this might be "shadowLayer" instead
    "$type": "shadow",
    "$value": {
      "color": {
        // in our case it's "$value": {"$ref": "color.accent"} instead
        "$value": "{color.accent}", 
        "$transforms": [
[...]

PR #246, opened 2024-08-17
Add layered shadow support, add support for inset shadows (non-breaking changes)

drwpow:
## Summary

This proposes the following changes to `$type: shadow` tokens:

- Allows arrays for `$value`
- Adds `inset` (boolean) property

## Reasoning

This PR proposes that the following discussions be resolved like so:

- #100: ✅ **Accepted** (all proposals that I saw)

## Pros

- Adds layered shadow support ([example](https://shadows.brumm.af/)) without breaking existing shadow tokens
- Adds support for inset shadows, also without breaking backwards compatibility

## Cons

- Arrays vs non-arrays adds a minor bit of tooling annoyance, but nothing significant

## Alternatives 

- We could just enforce “always arrays,” but IMO that’s not necessary to put that burden on the people writing JSON. The spec allows for other “syntactic sugar” and this seems like a good <abbr title="Quality of Life">QoL</abbr> addition
- We could make `inset` required, at the expense of breaking backward compatibility
- Instead of `"inset": [boolean]`, we could have something like `"position": "inner"` (string enum), but I can’t think of any other value this would be (what is there other than “inside” and “outside”?). Seems like designing for something no one has asked for.

## Notes

- Other token types such as gradients and cubic béziers already have arrays for values, so this doesn’t introduce inconsistencies in the design

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -96,7 +96,7 @@ Represents the style applied to lines or borders. The `$type` property MUST be s
 - an object value as defined in the corresponding section below
 
 <div class="issue" data-number="98" title="Stroke style type feedback">
-  Is the current specification for stroke styles fit for purpose? Does it need more sub-values (e.g. equivalents to SVG's `stroke-linejoin`, `stroke-miterlimit` and `stroke-dashoffset` attributes)? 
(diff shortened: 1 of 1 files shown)


kaelig (approved):

kaelig:
This proposal looks good to me, we'll make adjustments to the color values separately shortly.

PR #251, opened 2024-09-13
Update README.md - Adding Liferay to Open Source Companies List

marcoscv-work:
As an open-source Product, Liferay has been following the group activities and I'm personally part of it, we are totally aligned to DTCG and we are already working on aligning our token definitions.

We would like to be part of it more officially and show our support by appearing on the list of Open Source Companies and Projects represented in the DTCG.


kaelig (approved):

kaelig:
Thank you for your support!

PR #254, opened 2024-10-24
Fix design token heading levels

drwpow:
## Changes

Fixes an unintentional heading level issue under “Design token”

**Before**

Section 5.2 seems to suggest that the following headings are underneath it (which would mirror Groups):

before

**After**

Section 5.2 makes more sense with sub-headings (and makes Tokens and Groups consistent)

after

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -84,7 +84,7 @@ Furthermore, due to the syntax used for [token aliases](#aliases-references) the
 
 While `$value` is the only required property for a token, a number of additional properties MAY be added:
 
-## Description
+### Description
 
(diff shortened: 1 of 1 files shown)


technical-reports/format/design-token.md
```diff
@@ -178,5 +178,3 @@ Tool vendors are encouraged to publicly share specifications of their extension
 <p class="ednote" title="Extensions section">
   The extensions section is not limited to vendors. All token users can add additional data in this section for their own purposes.
 </p>
-
-## More token properties TBC
```
drwpow, line 182: Opinionated: I just removed this; other properties can be added here as normal revisions to the spec

c1rrus (approved):
Good spot. Thanks for fixing this

Issue #161, opened 2022-07-27
Status object: simplified audit/decision history


hereinthehive:
This is something I'd been thinking about before joining. A state/status object feels extensible and more descriptive. I'd also argue that the deleted state is valid as you might want to capture _why_ a value was deleted.

With concepts like this, perhaps we're touching on whether tokens should have their own version of an ADR so there's an audit/history around choices/decisions made?

```tsx
{
  "colors": {
    "action": {
      "$value": "#ff0000",
      "$type": "color",
      "$status": [{
        "$state": "deprecated",
        "$description": "We'll be replacing this soon, please check out how the replacement value works for you",
        "$value": "#00ff00",
        "$timestamp": "20220707111111",
        "$note": "We changed the color based on user research",
        "$toolstamp": "Figma",
        "$author": {
          "$name": "Dan Donald",
          "$email": "dan@zeroheight.com"
        }
      },
      {
        "$state": "created",
        "$timestamp": "20220128093348",
        "$author": {
          "$name": "Dan Donald",
          "$email": "dan@zeroheight.com"
        }
      }]
    }
  }
}
```
The rationale here is that as tokens when used to their full potential are a real business asset, it's useful to know when and why a decision was made. Seeing a simple history allows us to have context for a value but also allow any given tool to write to it and have that shared understanding in a way we haven't to date. Arguably the notion of 'toolstamp' is a naive way of suggesting there should be a reference to what tool was used as part of the audit.

Really curious to hear your thoughts!

> I think having a `$state` property will be really beneficial to the spec as well.
> 
> In our implementation, we have a `state` property that can take one of the following states:
> 
> `Experimental → Active → Deprecated → Sunset → Deleted`
> 
> * **Experimental**: For introducing tokens for internal validation and trail purposes
> * **Active**: This token is currently actively being used. Tooling will help to ensure tokens are used.
> * **Deprecated**: This token has been flagged for future deletion and should no longer be used.
> * **Sunset**: This token has been marked for upcoming deletion and strictly should not be used.
> * **Deleted (hard)**: This token has been fully removed from the tokens API
> 
> Our tooling leverages this along with a `replacement` property to perform automatic or controlled migrations via tooling such as `eslint`, `stylelint`, `figma` etc.
> 
> For example, a `deprecated` token with a `replacement` can be warned against and autofixed by eslint. A `sunset` token will produce an error and can be autofixed and so on.
> 
> ```tsx
> {
>   link: {
>     pressed: {
[...]

Issue #118, opened 2022-03-20
$deprecated property?


romainmenke:
When a specific design token should no longer be used the current schema only gives the option to remove it.

This can easily block consumption of future changes until all uses have caught up with the removal.

This is a communication issue.
Design token files will eventually have breaking content changes.

_This can be partly handled by using a distribution method that supports some form of versioning.
That would allow consumers to stay on a version they are compatible with until they are ready to update. This is outside the scope of my request._

Having a `$deprecated` field would allow tools to give warnings and guidance about upcoming changes.

```json
{
  "Button background": {
    "$value": "#777777",
    "$description": "The background color for buttons in their normal state.",
    "$deprecated": "Button styles have been moved to a separate design tokens file. Import 'buttons.tokens'"
  }
}
```

CITguy:
Why not a `$status` field instead?  

I can see this being useful to describe a token throughout the entirety of its lifespan (not just during deprecation).

Could start with just a few supported values...

- `experimental`
- `stable` (default)
- `deprecated`


_crude, partial JSON Schema doc to illustrate my idea..._
```javascript
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "token",
  "type": "object",
  "properties": {
    "$value": {},
    "$type": { "type": "string" },
    "$extensions": { "type": "object" },
    "$state": {
      "type": "string",
      "default": "available",
      "enum": [
        "experimental",
        "available",
        "deprecated"
      ]
    }
  },
  "required": ["$value"]
}
```

romainmenke:
I like this! But it will only work as long as no two statuses need to be assigned at the same time.

- Maybe `$state` can take an array?
- Maybe `$state` is an object? (`{ "$state": { "experimental": true, "deprecated": true } }`)

Something can be both deprecated and experimental.

danieldelcore:
I think having a `$state` property will be really beneficial to the spec as well. 

In our implementation, we have a `state` property that can take one of the following states: 

`Experimental → Active → Deprecated → Sunset → Deleted`

- Experimental: For introducing tokens for internal validation and trail purposes
- Active: This token is currently actively being used. Tooling will help to ensure tokens are used.
- Deprecated: This token has been flagged for future deletion and should no longer be used.
- Sunset: This token has been marked for upcoming deletion and strictly should not be used.
- Deleted (hard): This token has been fully removed from the tokens API

[...]

Issue #253, opened 2024-09-28
ambiguity around nested vs. flat structure


markacianfrani:
[6. Groups](https://tr.designtokens.org/format/#groups) states:
>A file MAY contain many tokens and they MAY be nested arbitrarily in groups like so:

```
{
  "token uno": {
    "$value": "#111111",
    "$type": "color"
  },
  "token group": {
    "token dos": {
      "$value": "2rem",
      "$type": "dimension"
    },
    "nested token group": {
      "token tres": {
        "$value": 33,
        "$type": "number"
      },
      "Token cuatro": {
        "$value": 444,
        "$type": "fontWeight"
      }
    }
  }
}
```

But [6.2.1](https://tr.designtokens.org/format/#file-authoring-organization) also states:
>Groups let token file authors better organize their token files. Related tokens can be nested into groups to align with the team's naming conventions and/or mental model. When manually authoring files, using groups is also less verbose than a flat list of tokens with repeating prefixes.
```
  "brand": {
    "color": {
      "$type": "color",
      "acid green": {
        "$value": "#00ff66"
      }
    },
```

...is likely to be more convenient to type and, arguably, easier to read, than:

```
  "brand-color-acid-green": {
    "$value": "#00ff66",
    "$type": "color"
  },
```

Does that mean the two are supposed to represent the same thing? Because the first example represents a token "acid green" and the second example represents a different token "brand-color-acid-green". 

This seems to suggest there are two different ways to parse a token file: nested or flat. Nested means I should consider the entire token path as part of the token name. Flat means I should assume that the token name is already flat and that any nested paths are just an arbitrary way of grouping items. 

It's very confusing to see that:
```
    "nested token group": {
      "token tres": {
        "$value": 33,
        "$type": "number"
      }
```
represents a token named `token-tres` and NOT `nested-token-group-token-tres` while also seeing that 

```
  "brand": {
    "color": {
      "$type": "color",
      "acid green": {
        "$value": "#00ff66"
      }
    },
```
represents a token named `brand-color-acid-green`. 

I feel like it should be possible to consume a .token file and be able to output a list of all the available design tokens using only the format. Would it be possible to add more clarification around these concepts?

PR #255, opened 2024-10-24
Proposal: add $deprecated property

drwpow:
## Summary

This adds a new `$deprecated` property for both tokens and groups [as suggested by @romainmenke](https://github.com/design-tokens/community-group/issues/118).

## Reasoning

This PR basically accepts #118 as-proposed (the original proposal), while taking into account feedback from followup comments.

### Pros

- This mirrors prior art like [JSDoc deprecation](https://jsdoc.app/tags-deprecated) that is well-known and well-used
- Style dictionary also [has this concept](https://github.com/design-tokens/community-group/issues/118#issuecomment-2277926212) and it makes sense to codify in the spec
- This seemed to have fairly-universal support (only disagreements were minor syntax differences, but that also spawned [a separate proposal](https://github.com/design-tokens/community-group/issues/161) that can be evaluated independently)

### Cons

- Expanding `$deprecated` in the future would be a breaking change
- What constitutes “deprecation” is a little ambiguous in the spec as-written. But the hope is that it’s no more ambiguous than using [this exact term elsewhere](https://jsdoc.app/tags-deprecated).
- What “deprecation” means is up to tools to decide (e.g. whether or not to generate output from those tokens, and/or throw warnings/errors)

## Alternatives

- Adding a pair property (e.g. `$deprecated` / `$deprecatedMessage`) isn’t a design pattern that currently exists in the spec. Pair properties can also be antipatterns as they introduce validation complexity (e.g. what if `$deprecatedMessage` is set but `$deprecated` isn’t—is that valid or invalid? and if valid, what behavior is inferred?)
- The object syntax (e.g. `$deprecated.deprecated` / `$deprecated.message`) wasn’t proposed here because it adds unnecessary boilerplate.
- A more complex [status object](https://github.com/design-tokens/community-group/issues/161) would be a separate proposal because that encroaches on opinionated versioning (and versioning won’t map 1:1 to existing tokens)
  - This proposal assumes that any implementation of status/versioning won’t conflict with `$deprecated`. Or in areas where they potentially overlap, will be complementary rather than conflicting.
  - In other words, acceptance of this does NOT mean refusal of the Status Object proposal

## Notes
- This assumes **aliases may be deprecated too** (even if the tokens they point to don’t). Should this be outlined explicitly?
- It is the author’s opinion that deprecation is not directly related to versioning. Versioning is a more complex opinion on what constitutes “breaking changes” from non-, e.g. [semver](https://semver.org/). Though versioning _may_ include deprecation and/or removal of tokens, it’s so much more. Conversely, the act of deprecation is common enough it affects every design system, even those that may not have defined versioning strategies. So whether or not this spec defines an opinionated versioning strategy, shouldn’t affect this proposal.

### DIFF
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -181,7 +181,7 @@ Tool vendors are encouraged to publicly share specifications of their extension 
 
 ### Deprecated
 
-The optional **`$deprecated`** property is a boolean or string where tools MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
+The **`$deprecated`** property MAY be used to mark a token as deprecated, and optionally explain the reason. Reasons to deprecate tokens include but are not limited to the following:
 
 - A future update to the design system will remove this token
 - Removing the token now may break existing support
@@ -199,7 +199,7 @@ The optional **`$deprecated`** property is a boolean or string where tools MAY s
   "Button focus": {
     "$value": "#70c0ff",
     "$type": "color",
-    "$deprecated": "Please use {button.activeBorder} instead."
+    "$deprecated": "Please use the border style for active buttons instead."
   }
 }
 ```
@@ -211,3 +211,27 @@ The optional **`$deprecated`** property is a boolean or string where tools MAY s
 | `true`   | This token is deprecated (no explanation provided).         |
 | `String` | This token is deprecated AND this is an explanation.        |
 | `false`  | This token is NOT deprecated (may override group defaults). |
+
+Tool makers MAY augment the string when it contains aliases such as the one given as an example. A tool could potentially resolve the token, and link to docs, code, or render a visual representation of it and link to the new token inside a UI. For example, “Please use {button.activeBorder} instead“ could be output in JS as:
+
+<aside class="example">
+
+```js
+/**
+ * @deprecated Please use {@link file://./my-tokens.js:85 button.activeBorder} instead.
+ */
+```
+
+</aside>
+
+Or
+
+<aside class="example">
+
+```js
+/**
+ * @deprecated Please use {@link https://docs.ds/tokens/#button-activeborder button.activeBorder} instead.
+ */
+```
+
+</aside>
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -178,3 +178,36 @@ Tool vendors are encouraged to publicly share specifications of their extension 
 <p class="ednote" title="Extensions section">
   The extensions section is not limited to vendors. All token users can add additional data in this section for their own purposes.
 </p>
+
+### Deprecated
+
+The optional **`$deprecated`** property is a boolean or string where tooks MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
+
+- A future update to the design system will remove this token
+- Removing the token now may break existing support
+- This token is discouraged from future use
+
+<aside class="example">
+
+```json
+{
+  "Button background": {
+    "$value": "#777777",
+    "$type": "color",
+    "$deprecated": true
+  },
+  "Button focus": {
+    "$value": "#70c0ff",
+    "$type": "color",
+    "$deprecated": "Please use {button.activeBorder} instead."
+  }
+}
+```
+
+</aside>
+
+| Value    | Explanation                                                 |
+| :------- | :---------------------------------------------------------- |
+| `true`   | This token is deprecated (no explanation provided).         |
+| `String` | This token is deprecated AND this is an explanation.        |
+| `false`  | This token is NOT deprecated (may override group defaults). |
--- a/technical-reports/format/design-token.md
+++ b/technical-reports/format/design-token.md
@@ -181,7 +181,7 @@ Tool vendors are encouraged to publicly share specifications of their extension 
 
 ### Deprecated
 
-The optional **`$deprecated`** property is a boolean or string where tooks MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
+The optional **`$deprecated`** property is a boolean or string where tools MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
 
 - A future update to the design system will remove this token
 - Removing the token now may break existing support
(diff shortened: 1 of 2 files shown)


technical-reports/format/design-token.md
```diff
@@ -179,4 +179,35 @@ Tool vendors are encouraged to publicly share specifications of their extension
   The extensions section is not limited to vendors. All token users can add additional data in this section for their own purposes.
 </p>
 
-## More token properties TBC
+### Deprecated
+
+The optional **`$deprecated`** property is a boolean or string where tooks MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
```
drwpow, line 184: cc @c1rrus any nitpicking of language here would be greatly appreciated 🙏. Happy to rephrase this to match the existing language (or if you wanted to do a pass on it that would be fine, too!)

romainmenke (approved):
Thank you. LGTM 👍

technical-reports/format/design-token.md
c1rrus, line 184: Minor typo

```suggestion
The optional **`$deprecated`** property is a boolean or string where tools MAY specify deprecated tokens. A token MAY be marked deprecated in any of the following scenarios:
```

technical-reports/format/groups.md
```diff
@@ -206,6 +206,36 @@ In this example, the "brand" group has 2 extensions: `org.example.tool-a` and `o
 
 </aside>
 
+### Deprecated
+
+The optional **`$deprecated`** property is a boolean or string where tooks MAY specify all tokens within the group are deprecated. A token MAY be marked deprecated in any of the following scenarios:
```
c1rrus, line 211: Same typo here.

```suggestion
The optional **`$deprecated`** property is a boolean or string where tools MAY specify all tokens within the group are deprecated. A token MAY be marked deprecated in any of the following scenarios:
```

(aside: Maybe we should consider having a single chapter for all token and/or group properties, since many exist for both. Each property could specify if it can be used on tokens, groups or both. That could avoid a lot of the duplication and copy-pasting we need to do today. Obvs. that would need to be a separate PR)

c1rrus (changes requested):
Spotted a couple of teeny typos. But otherwise this looks good. Happy to approve once they get fixed.

technical-reports/format/groups.md
drwpow: >(aside: Maybe we should consider having a single chapter for all token and/or group properties, since many exist for both.

That would make sense to me. It could also possibly clarify [some in-the-weeds confusion around parsing token vs group nodes](https://github.com/design-tokens/community-group/issues/253)

c1rrus (approved):
Nice one. Thanks for soing this @drwpow!

drwpow:
Thanks! Will give other editors a chance to review and leave feedback over the next week or 2. If there’s no objection, we’ll merge!

But also, for the other proposals moving forward, generally the editors express the same opinions found in the comments, so the fact it’s being proposed in the first place is usually positive 🙂

technical-reports/format/design-token.md
```diff
@@ -178,3 +178,36 @@ Tool vendors are encouraged to publicly share specifications of their extension
...
+{
+  "Button background": {
+    "$value": "#777777",
+    "$type": "color",
+    "$deprecated": true
+  },
+  "Button focus": {
+    "$value": "#70c0ff",
+    "$type": "color",
+    "$deprecated": "Please use {button.activeBorder} instead."
```
CITguy, line 202: Would downstream tooling need to convert `{button.activeBorder}` to a language-specific name?  This syntax is typically used for aliasing, and there's no clear instruction on:

1. Should this alias be resolved?
    1.  If so, what value should it resolve to?

drwpow: Yeah that’s a good question. I could see this being converted to, say, JSDoc or a CSS comment like so:

```tsx
  /** @deprecated Please use {button.activeBorder} instead */
  'Button focus': '#70c0ff',
```

In which case, would that be clear to people? The simplest interpretation is “the designers/dev write this message in a way that is clearest to their team” with no official guidance. I’m generally against parsing strings in any way, but yeah to your point there is a good question of “could this be improved?”

Nothing comes to mind, because in many cases there is no 1:1 replacement; it’s often a mixture of things. Perhaps we just adjust this example to not mention other tokens to avoid confusion?

technical-reports/format/groups.md
```diff
@@ -206,6 +206,36 @@ In this example, the "brand" group has 2 extensions: `org.example.tool-a` and `o
...
+- A future update to the design system will remove this token
+- Removing the token now may break existing support
+- This token is discouraged from future use
+
+<aside class="example">
+
+```json
+{
+  "Button": {
+    "$deprecated": "Please use {action.*} tokens instead.",
```
CITguy, line 222: (see [similar comment](https://github.com/design-tokens/community-group/pull/255/files#r1819400982))

technical-reports/format/groups.md
```diff
@@ -206,6 +206,36 @@ In this example, the "brand" group has 2 extensions: `org.example.tool-a` and `o
...
+```
+
+</aside>
+
+In the context of a group, adding `$deprecated` will apply to all tokens within that group, unless a token explicitly sets a value of `false`. Any value provided by a token will override the group default.
+
+| Value    | Explanation                                                 |
+| :------- | :---------------------------------------------------------- |
+| `true`   | This token is deprecated (no explanation provided).         |
+| `String` | This token is deprecated AND this is an explanation.        |
```
kaelig, line 236: Perhaps it would be useful to mention that tool makers MAY augment the string when it contains aliases such as the one given as an example:

```
Please use {button.activeBorder} instead
```

lukasoppermann:
@drwpow good to merge?

drwpow:
> @drwpow good to merge?

Almost! We had been waiting to see if one group had feedback. Will ping them again and if not, will merge!

drwpow:
> To confirm, there is neither a necessity nor a spec-defined way to confirm or extract an alternate token suggestion inside the deprecated tag, toolmakers can make their own assumptions on if and how to programmatically determine it?

Correct. Updated the wording per @kaelig’s suggestion to describe this a bit more. By default we don’t have aliases in the examples anymore outside the section that describes “it’s up to tools to determine how they do/don’t handle parsing the `$deprecated` string.

drwpow:
We’ve waited for additional parties to give input / feedback but haven’t gotten any pushback or corrections. Merging!

PR #256, opened 2024-10-28
Shadow token examples dimension format

mark-tomlinson-dev:
# Fix Shadow Token Examples to Match Specification

Updated both the `layered-shadow` and `inner-shadow` examples to use the correct dimension object format:
- e.g. `"24px"` → `{ "value": 24, "unit": "px" }`

This ensures the examples are consistent with the dimension value specification as outlined in section 8.2.


drwpow (approved):
Great catch, thank you! I think this accidentally got skipped because both proposals were open for feedback at once, and didn’t get rebased properly.

Issue #137, opened 2022-06-14
Legacy color


o-t-w:
From the part of the spec about color types: “Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The $type property MUST be set to the string color. The value MUST be a string containing a hex triplet/quartet including the preceding # character. To support other color spaces, such as HSL, export tools SHOULD convert color tokens to the equivalent value as needed.”

Chrome is currently working in implementing LCH, LAB, OKLCH and OKLAB color. Safari already supports them. Although support is currently rarer among design tools it still seems sad and backwards-looking to standardise and require a legacy color format.

It is impossible to transform a hex color into a modern color. It is overwhelmingly likely that the modern color formats will become ubiquitous on the web within the next year or two. I'm sure design tools will quickly follow along. 

My suggestion would be to standardize on OKLCH already and have design tools that don't yet support modern color transform the color into hex. 

I'm already using LAB and LCH on a large web application with hex fallbacks. Standardizing on hex would make design tokens unusable for me.

romainmenke:
The only format that makes sense to me is something like this :

```json
{
  "colorSpace": "sRGB",
  "channels": [0.1, 0.1, 0.1],
  "alpha": 1.0
}
```

The different representations of a color space (hsl for rgb, lch for lab, ...) are less important in a transfer format like design tokens.

sebfriedrich:
Fully agree @o-t-w , I don't use wide gamut color but have my thoughts sorted on how I would want to have it.

For the ones who aren't convinced here: What Chrome and other browsers implement there right now dates back 90 years from now to color theory research about human perception. Also design systems are not meant to cover a subset of what CSS can help one to build, it is the other way around. It is also worth noting, that a design system shall help manage media breaks more reliability than without such a system in place. Delegating color conversion, which shall be assumed lossy by default, to a random tool at sort of runtime is accepting that applying color via tokens will have different and unforseeable deviations.

CieLCh or similar might be a good start for thinking. Ultimately I would hope for the ability to describe any custom color space by an 3x3 conversion matrix and for better usability hard code the most common ones as named presets to use out-of-the-box.

The work of Bruce Lindbloom might be a good read to get into the math of the topic: http://www.brucelindbloom.com/

NateBaldwinDesign:
[...]

PR #257, opened 2024-11-08
Color format adjustments

resource11:
## Summary

This adds new `$colorSpace`, `colorProfile`, `coordinates`, and `fallback` properties for the color type `$value` object [as suggested by multiple discussions](https://github.com/design-tokens/community-group/issues/137).

## Reasoning

This PR takes into account feedback from followup comments with respect to loosening up the color type and gracefully handling legacy color scenarios.

### Pros

- What constitutes “color” is somewhat narrow in the spec as-written.The hope is that these additions will empower toolmakers and consumers to leverage modern color spaces, while supporting legacy color profiles depending on display needs.

### Cons

- Expanding `$color` in the future would be a breaking change, yet the release of the color format module will already be a breaking change.

## Alternatives

- The object syntax (e.g. `$components` / `$channels`) was modified to align with both [CSS Color Module Level 5](https://drafts.csswg.org/css-color-5/#intro) and [ColorJS](https://colorjs.io/docs/spaces) naming conventions for consistency.
  - This proposal assumes that the `color` module will provide expanded guidance on this format. Or in areas where they potentially overlap, differences will be complementary rather than conflicting.

## Notes
- The color first draft PR will be updated depending on feedback here.


technical-reports/color/color-type.md
```diff
@@ -29,7 +29,7 @@ For example, initially color tokens may be defined as such:
 
 ```json
 {
-  "Majestic magenta": {
+  "Hot pink": {
     "$type": "color",
     "$value": {
       "$hex": "#ff00ff"
```
drwpow, line 35: For this file, we should just define it in the `srgb` space, and we don’t need the  `hex` fallback:

```suggestion
      "colorSpace": "srgb",
      "channels": [1, 0, 1]
```

Here and below they should follow the same format

technical-reports/color/color-type.md
```diff
@@ -75,20 +75,20 @@ For example, initially color tokens may be defined as such:
 
 ```json
 {
-  "Majestic magenta": {
+  "Hot pink": {
     "$type": "color",
     "$value": {
-      "$hex": "#c44587",
+      "$hex": "#ff00ff",
       "$colorSpace": {
```
drwpow, line 82: Here, and below, these should align with the current spec:

```suggestion
      "colorSpace": "srgb",
      "channels": [1, 0, 1],
```

technical-reports/format/composite-types.md
```diff
@@ -423,7 +423,7 @@ Describes a gradient that is solid yellow for the first 2/3 and then fades to re
 {
   "brand-primary": {
     "$type": "color",
-    "$value": "#99ff66"
+    "$value": "#00ff66"
```
drwpow, line 426:
```suggestion
    "$value": {
      "colorSpace": "srgb",
      "channels": [0, 1, 0.4]
```

technical-reports/format/groups.md
```diff
@@ -80,7 +80,7 @@ Group keys without a dollar sign (`$`) prefix denote:
       "$description": "This is an example of a group containing a single token",
       "Token name": {
         "$value": "#000000",
```
drwpow, line 82: Here, and below, all colors should match the updated format:

```suggestion
        "$value": {
          "colorSpace": "srgb",
          "channels": [0, 0, 0]
        }
```

I think we don’t need `hex` on these, too, because we’re just specifying sRGB colors (however, if using a color gamut or bit depth higher than 8-bit sRGB, it could be provided).

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+Represents a color in the UI. The `$type` property MUST be set to the string `color`. The `$value` property MUST be an object string containing a string `colorSpace`, a numeric `coordinates` array, optional numeric `alpha`, string `colorProfile`, and string `fallback` to support legacy color spaces as needed.
+
+| Key            |      Type      | Required | Description                                                                                                                              |
+| :------------- | :------------: | :------: | :--------------------------------------------------------------------------------------------------------------------------------------- |
+| `colorSpace`   |    `string`    |    Y     | An string representing the color space.                                                                                                  |
+| `colorProfile` |    `string`    |          | String representing url of color profile.                                                                                                |
```
drwpow, line 20: I think we are not using colorProfile, and just aligning with CSS Color Module 4? `colorSpace` is a shorthand for space + profile + gamut, also borrowing from the same spec

```suggestion
```

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+Represents a color in the UI. The `$type` property MUST be set to the string `color`. The `$value` property MUST be an object string containing a string `colorSpace`, a numeric `coordinates` array, optional numeric `alpha`, string `colorProfile`, and string `fallback` to support legacy color spaces as needed.
+
+| Key            |      Type      | Required | Description                                                                                                                              |
+| :------------- | :------------: | :------: | :--------------------------------------------------------------------------------------------------------------------------------------- |
+| `colorSpace`   |    `string`    |    Y     | An string representing the color space.                                                                                                  |
```
drwpow, line 19: Note: we could probably link to all the valid values from CSS Color Module 4 (or possibly 5, if we want to extend to that)

```suggestion
| `colorSpace`   |    `string`    |    Y     | A string representing the color space.                                                                                                  |
```

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+| `colorProfile` |    `string`    |          | String representing url of color profile.                                                                                                |
+| `coordinates`  | `number array` |    Y     | An array of numeric values representing individual color coordinates.                                                                    |
```
drwpow, line 21: I think we were leaning toward “channels” being the term. Also just in the interest of being really specific, “an array” doesn’t specify the length of the list. Calling it a “tuple of 3 values” means that the values must always be 3—no more, no less. Just like any color in CSS, I think we should enforce that the length is always 3, no matter what.

```suggestion
| `channels`  | `[number, number, number]` |    Y     | Tuple of 3 numbers representing the 3 components of the color space. The allowed ranges depend on the `colorSpace` specified. |
```

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+| `alpha`        |    `number`    |          | An integer or floating-point value representing the numeric value.                                                                       |
```
drwpow, line 22: We should specify this MUST be in the range of `(0, 1)` (inclusive). `0` = 0%, `1` = 100%.

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+| `fallback`     |    `string`    |          | Unit of distance. Supported values: [HEX 6](https://www.w3.org/TR/css-color-4/#typedef-hex-color) including the preceding `#` character. |
```
drwpow, line 23:
```suggestion
| `hex`     |    `string`    |          | Fallback sRGB-8 color for colors using higher gamuts or higher bit depths. MUST be in the format of a [HEX 6](https://www.w3.org/TR/css-color-4/#typedef-hex-color) string, including the preceding `#` character. |
```

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
 ```json
 {
-  "Majestic magenta": {
-    "$value": "#ff00ff",
-    "$type": "color"
+  "Hot pink": {
+    "$type": "color",
+    "$value": {
+      "colorSpace": "display-p3",
+      "colorProfile": "http://example.org/display-p3.icc",
```
drwpow, line 35: Here and below, the format should match the updated syntax of `colorSpace` + `channels` + `alpha` (optional) + `hex` (optional)

```suggestion
     "channels": [0.92, 0.2, 0.97],
      "hex": "#ff00ff" // HEX 6 supported only
```

technical-reports/format/types.md
```diff
@@ -12,21 +12,42 @@ If an explicit type is set, but the value does not match the expected syntax the
...
+| `colorProfile` |    `string`    |          | String representing url of color profile.                                                                                                |
+| `coordinates`  | `number array` |    Y     | An array of numeric values representing individual color coordinates.                                                                    |
+| `alpha`        |    `number`    |          | An integer or floating-point value representing the numeric value.                                                                       |
+| `fallback`     |    `string`    |          | Unit of distance. Supported values: [HEX 6](https://www.w3.org/TR/css-color-4/#typedef-hex-color) including the preceding `#` character. |
 
 For example, initially the color tokens MAY be defined as such:
 
 <aside class="example">
 
 ```json
```
drwpow, line 29: Suggestion: if using comments, `jsonc` (“JSON with **C**omments) won’t show visible errors in syntax highlighting

````suggestion
```jsonc
````

**Review by drwpow — COMMENTED** [review]
Thanks for updating this! Added some comments around updating the examples to match the format we last talked about in meetings. Note that small matters of `channel` vs `components` we can vote on / decide later! But I think it would be helpful to at least have all the examples be internally-consistent, and matching with the CSS Color Module 4 style of just using `colorSpace` to refer to space + profile + gamut.

File: technical-reports/format/types.md (hunk 7 above)
Comment (resource11) [reply technical-reports/format/types.md:21]: Yes, we were leaning towards that! 

The terms `component` and `channel` were used in different capacities in the W3C CSS Color Level docs, sometimes interchangeably. I agree that we should standardize terminology and clarify what this means.

For example in the [CSS Color Level 4](https://www.w3.org/TR/css-color-4/#color-type) color type definition it reads:

> Colors in CSS are represented as a list of color **components**, also **sometimes called “channels**”, representing axises in the color space. Each channel has a minimum and maximum value, and can take any value between those two. Additionally, every color is accompanied by an **alpha component**, indicating how transparent it is, and thus how much of the backdrop one can see through the color.

The [CSS Color Level 5](https://www.w3.org/TR/css-color-5/#intro) spec builds upon Level 4, yet the terminology is still interchangeable.

In [ColorJS](url), the term "coordinates" could be shorthand for component coordinate or channel coordinate if I assume correctly (need to confirm).

Happy to move back to `channels` for now. 

Also, I suggest we adjust wording to include `none` along with numbers in the allowable channel value types, to account for [4.4“Missing” Color Components and the none Keyword](https://www.w3.org/TR/css-color-4/#missing) support.

Comment (drwpow-figma) [reply technical-reports/format/types.md:21]: > Also, I suggest we adjust wording to include `none` along with numbers in the allowable channel value types, to account for [4.4“Missing” Color Components and the none Keyword](https://www.w3.org/TR/css-color-4/#missing) support.

Oh smart! I like the idea of aligning with all that great thinking as much as possible

File: technical-reports/format/types.md (hunk 10 above)
Comment (resource11) [reply technical-reports/format/types.md:35]: Thanks for this suggestion! I should clarify why the optional `colorProfile` is listed here. I was thinking of the cases for print where a specific color profile should be associated with the color space for conversion. Perhaps we do a separate callout for this and use CMYK as the color space example.

File: technical-reports/color/color-type.md (hunk 1 above)
Comment (resource11) [reply technical-reports/color/color-type.md:35]: Thanks for the feedback! No changes have been added to this file yet. Once we get alignment on the rest, we can make adjustments throughout here.

File: technical-reports/color/color-type.md (hunk 2 above)
Comment (resource11) [reply technical-reports/color/color-type.md:82]: See above comment

File: technical-reports/format/types.md (hunk 5 above)
Comment (resource11) [reply technical-reports/format/types.md:20]: In [CSS Color Module 4](https://drafts.csswg.org/css-color-4/#terminology) color terminology section, here's the definition of color space:

> A color space is an organization of colors with respect to an underlying colorimetric model, such that there is a clear, objectively-measurable meaning for any color in that color space. This also means that the same color can be expressed in multiple color spaces, or transformed from one color space to another, while still looking the same.

[Color profile](https://drafts.csswg.org/css-color-5/#at-profile) is different, it's about using a particular ICC color profile for device specific implementations (print, for example). For example, you can use the CMYK color space and choose more than one color profile to implement those colors.

File: technical-reports/format/types.md (hunk 7 above)
Comment (resource11) [reply technical-reports/format/types.md:21]: Yes, there have been interest in this support in the legacy color thread!

File: technical-reports/format/types.md (hunk 8 above)
Comment (resource11) [reply technical-reports/format/types.md:22]: Can do. Where should this sentence occur?

File: technical-reports/format/types.md (hunk 5 above)
Comment (drwpow) [reply technical-reports/format/types.md]: You’re 100% right, and yes I do see us needing this. As we talked about earlier today, perhaps we limit this initial PR to CSS Color Module 4 where:

- only `colorSpace` is used
- `colorSpace` is limited to the allowed values in CSS Color Module 4 (`srgb`, `srgb-linear`, `display-p3`, etc.)

Then, we open a separate PR to expand the format to include CSS Color Module 5’s concept of `@color-profile`. That way we can get feedback on that separately. Not being prescriptive, but if we copy CSS Color Module 5’s mental model, I could see it introducing a new `$type: 'colorProfile'` token type that maps 1:1 to `@color-profile` (perhaps; maybe we think of another pattern!)

File: technical-reports/format/types.md (hunk 8 above)
Comment (drwpow) [reply technical-reports/format/types.md]: Anywhere in the `alpha` description

File: technical-reports/format/composite-types.md (hunk 3 above)
Comment (drwpow) [reply technical-reports/format/composite-types.md:426]: OK! Yes breaking out changes to gradient as a separate PR would be easier to review! This PR shouldn’t introduce any changes to gradient. Because, yes, gradients do use colors, that doesn’t mean gradients change necessarily. Then we can review those unique changes easier

File: technical-reports/format/types.md (hunk 7 above)
Comment (drwpow) [reply technical-reports/format/types.md]: Oh also I think as I said in another comment, maybe we like `components` better and that’s fine with me! I don’t have any strong opinions. We could decide on the exact words at the last minute without it changing the proposal too much.

File: technical-reports/color/color-type.md (hunk 1 above)
Comment (drwpow) [reply technical-reports/color/color-type.md:35]: Just a suggestion: it is easier to review if this PR made everything consistent in all files, since that has to be done one way or another. Otherwise it ends up in a state of confusion not being able to tell where the proposal starts and where it ends. Would it be OK to just make everything a consistent format so it’s clear to reviewers?

**Review by drwpow — APPROVED** [review]
As stated in some of the comments, I understand that there will be multiple PRs to bring the larger color format PR up to speed.

I would still lean towards Color Module 4 (and omit `colorProfile` and CMYK spaces) just for the purposes of this proposal, and we could discuss Color Module 5 in a followup, as an incremental addition. Hopefully by narrowing focus to Color Module 4 parity we could review and ship this faster.

Thanks again for doing all the work!

File: technical-reports/format/types.md (hunk 7 above)
Comment (romainmenke) [reply technical-reports/format/types.md]: > Calling it a “tuple of 3 values” means that the values must always be 3—no more, no less.

This should be an array of arbitrary length.
CMYK for example has 4 channels.
Other print profiles might have many more for spot colors and varnishes.

Defining these as arrays now means that implementers are less likely to make assumptions that these will always have a length of 3.

File: technical-reports/format/index.html [hunk 12]
```diff
@@ -19,20 +19,11 @@
...
-          { name: 'Donna Vitan', url: 'https://www.donnavitan.com/' },
-          { name: 'Drew Powers', url: 'https://pow.rs/' },
+          { name: 'Donna Vitan', url: 'http://twitter.com/donnavitan' },
           { name: 'James Nash', url: 'https://cirrus.twiddles.com/' },
-          { name: 'Kathleen McMahon', url: 'https://kathleenmcmahon.dev' },
           { name: 'Kevin Powell', url: 'https://twitter.com/kevinmpowell' },
           { name: 'Louis Chenais', url: 'https://twitter.com/chuckn0risk' },
-          { name: 'Mike Kamminga', url: 'https://twitter.com/mikekamminga' },
-          { name: 'Matthew Ström', url: 'https://matthewstrom.com/' },
         ],
```
kaelig, line 27: What happened there?

kaelig (changes requested):

technical-reports/format/index.html
resource11: Good question! Although we're reviewing the the changes from this branch in the [updated color-first-draft branch](https://github.com/design-tokens/community-group/pull/266) now, it makes sense to close the loop on this comment, @kaelig. Here's the gist:

In August, after one of our format meetings with @drwpow and yourself, I made a [color-first-branch draft commit](https://github.com/design-tokens/community-group/commit/c58ca34ada13f3d072a81cd72b7c7fbcfe3a3d20#diff-3efeb356792720be185f3d92610cd1a227b183088864b52b4da5d68a807713f2) to update the editors list from this:

```json
editors: [
          { name: 'Daniel Banks', url: 'https://twitter.com/dbanksDesign' },
          { name: 'Donna Vitan', url: 'http://twitter.com/donnavitan' },
          { name: 'James Nash', url: 'https://cirrus.twiddles.com/' },
          { name: 'Kevin Powell', url: 'https://twitter.com/kevinmpowell' },
          { name: 'Louis Chenais', url: 'https://bsky.app/profile/lucho.cool' },
        ],
```

to this:
```json
editors: [
          { name: 'Adekunle Oduye', url: 'https://www.adekunleoduye.com/' },
          {
            name: 'Ayesha Mazrana (Mazumdar)',
            url: 'https://twitter.com/AyeshaKMaz',
          },
          { name: 'Daniel Banks', url: 'https://twitter.com/dbanksDesign' },
          { name: 'Donna Vitan', url: 'https://www.donnavitan.com/' },
          { name: 'Drew Powers', url: 'https://pow.rs/' },
          { name: 'James Nash', url: 'https://cirrus.twiddles.com/' },
          { name: 'Kathleen McMahon', url: 'https://kathleenmcmahon.dev' },
          { name: 'Kevin Powell', url: 'https://twitter.com/kevinmpowell' },
          { name: 'Louis Chenais', url: 'https://twitter.com/chuckn0risk' },
          { name: 'Mike Kamminga', url: 'https://twitter.com/mikekamminga' },
          { name: 'Matthew Ström', url: 'https://matthewstrom.com/' },
        ],
```

When we reviewed this in the next meeting, you asked me to revert it and split out this change for separation-of-concerns purposes. So I did. 😄 

This reversion showed up here in this separate branch.

resource11:
Closing, these changes are reflected in #266.

Issue #166, opened 2022-08-30
Can alias tokens reference tokens in another file?


TravisSpomer:
Theming (#2) is definitely still a big unresolved issue, but let's say for now you have three token files all in this format: `global.json`, `light.json`, and `dark.json`.

#### global.json
```json
{ "black": { "$type": "color", "$value": "#000000" },
  "white": { "$type": "color", "$value": "#ffffff" } }
```

#### light.json
```json
{ "foreground": { "$type": "color", "$value": "{black}" } }
```

#### dark.json
```json
{ "foreground": { "$type": "color", "$value": "{white}" } }
```

You combine them with something like this:

```bash
theoretical-token-tool --in global.json --in light.json --out light.css
theoretical-token-tool --in global.json --in dark.json --out dark.css
```

On their own, `light.json` and `dark.json` are useless; they require `global.json` to make sense. My open-ended question is: are `light.json` and `dark.json` *invalid* DTCG-format token files that should not be accepted by any tool because their references can't be resolved on their own? Or are they still *valid*, but just can't be *used* without `global.json`?

I've been assuming all this time that all three files are valid, but I didn't see anything in the spec that says either way.

* **If they're valid:** The spec should explicitly say that it's valid to have a token file that references tokens that don't exist in that file.
* **If they're not valid:** One of the following will happen:
  1. Tools like `theoretical-token-tool` will decide to accept those invalid files anyway, which will end up hurting interoperability
  2. Or people who need multiple themes will have to put all of their themes in a single file and give a prefix like `light.` or `dark.` to all of their themed tokens, and then strip that prefix later when exporting to another format like CSS
  3. Or people who need multiple themes have to include all of the tokens from `global.json` in both `light.json` and `dark.json`

I'm not sure if we have to solve theming for v1 of this spec, but it seems like we do at least have to define this. All of those options have downsides, and if the spec doesn't specify the behavior, I think we'll get several downsides all at once as each tool interprets things a little differently.

c1rrus:
Thanks for raising this. I completely agree that this is something the spec needs to be more explicit about.

I think this relates closely to the discussion in #123 - i.e. should the spec be explicit about how multiple token files are processed and their contents are merged.

As I proposed there, I wonder whether we ought to also think about some kind of include or import mechanism. While we'd still need to figure out exactly how those included files are processed and merged, it could provide a more convenient experience for end users.

[...]

PR #259, opened 2025-01-11
Proposal: breaking changes to Aliases (JSON Schema $ref)s

drwpow:
## Summary

This makes a breaking change to aliases, changing the syntax:

```diff
- { "$value": "{color.blue.01}" }
+ { "$value": { "$ref": "#/color/blue/01" } }
```

This provides a solution for #166 as well as makes the implementation for the upcoming Resolver Spec that addresses #210.

## Reasoning

DTCG users have kept their files separately for a long time, and have asked for the ability to reference tokens in other files (#166). Since we were already borrowing from [JSON Schema](https://json-schema.org/draft/2020-12/json-schema-core#name-schema-references) in places (primarily the `$` character to mark reserved keys), this brings JSON Schema’s `$ref` keyword to the concept of token aliasing.

The `$ref` keyword comes both from JSON Schema and the [OpenAPI specification](https://spec.openapis.org/oas/latest.html) (previously _Swagger_), and has been in use for over a decade.

For the DTCG format, it is a way to keep all the functionality of aliases while extending it for future needs, borrowing from prior art where this syntax is known, used, and well-defined. For simplicity, this proposal aims to **replace** the previous alias syntax (which will be covered below).

### Pros

- This greatly expands what’s possible with the DTCG format, now that any part of the document may be reused.
  - For example, aliased tokens could now even reuse parts of `$extensions` if they wanted to, while selectively applying overrides
  - Because it allows overrides, you can reuse more DTCG syntax than you could before, even sub-schemas in remote documents (e.g. extending public design systems)
- While this is flagged as a breaking _syntax_ change, it’s actually a fully-backwards-compatible upgrade to aliases. There is no known functionality lost from this change.
- Figma Styles and Variables use the `/` character in names, so now token names will map 1:1
  - _Are there other cases where this is true?_
- This is backwards-compatible for token names. Tokens can still use `/` and `#` characters in their names; they just have to be escaped with `~0` and `~1` respectively, according to the [spec](https://datatracker.ietf.org/doc/html/rfc6901#section-3):

   ```json
   {
     "my/token": { "$type": "number", "$value": 5 },
     "other-token": { "$ref": "#/my~1token" }
   }
   ```
- For DTCG parsers, this simplifies parsing/handling of aliases. Consider the old syntax:
   ```json
   {
     "font-1": { "$type": "fontFamily", "$value": "Inter" },
     "font-2": { "$type": "fontFamily", "$value": "{font-1}" }
   }
   ```

   A tool maker would have to do additional work figuring out if `$value` was an alias or not, because both present as strings. But with the `$ref` syntax, if it’s an object with a `$ref` key, it is only an alias, no string parsing necessary.
- Tool makers also benefit from over a decade of prior art and tooling for handling this syntax (even if they may not have used it directly before)

### Cons

- Given that aliases are core syntax, this is a disruptive change, not just for toolmakers but for consumers of the DTCG spec. We likely would want to talk about a deprecation strategy
- It does change the structure, converting a string node (`"{my.token.alias"}`) to an object (`{ "$ref": "#/my/token/alias" }`). Again, though long-term it will be easier to work with, short-term will impose migration pains.

### Alternatives

- Those familiar with JSON Schema will also be familiar with its counterpart [`$defs`](https://json-schema.org/draft/2020-12/json-schema-core#name-schema-re-use-with-defs), which allows you to declare reusable parts of a document that can be `$ref`’d anywhere (but by default are ignored/not parsed). $defs are NOT being proposed here, because it doesn’t solve an immediate problem, and introduces more complexity than necessary (but could be an additive followup)
- An alternate idea is to keep _aliases_ the same `{color.blue.05}`, and introduce this new concept as a _reference_ (and distinguish between the terms). Aliases would just be the “legacy” way to declare token aliases. I didn’t initially propose this because:
  - There were no advantages I could find. Even the automatic `$type` inheritance is possible with `$ref`, but `$ref` carries more benefits
  - I couldn’t think of a sensible usecase of distinguishing “aliases” from “references” after reading this note in the [JSON Schema spec (2019-09](https://json-schema.org/draft/2019-09/draft-handrews-json-schema-02#rfc.appendix.B.2):
  
    > Attempting to remove all references and produce a single schema document does not, in all cases, produce a schema with identical behavior to the original form.

    In other words, even the concept of `$ref` feels spiritally identical to the existing DTCG alias syntax—there are times when you _do_ want to preserve and reference those values later, and they are significant in some ways (even including overrides).
    
  However, I think we could define a some sort of deprecation strategy where the old syntax is supported for some time before switching over.

## Notes

- This does introduce some confusion in syntax. For example, is the “official” token ID now `color.blue.05` or `color/blue/05`?
  - Further, does this mean `.` is allowed in token names again?
- As someone that’s worked extensively with JSON Schema syntax, it’s often not enough to say “follow the spec” because there are, like, dozens of conflicting versions that all have breaking changes. So I’m proposing specifically the [2020-12](https://json-schema.org/draft/2020-12/json-schema-core) version, in the case that tool makers run into one of these conflicts
  - Specifically, `2020-12` DOES allow `$ref` to have sibling keys (“overrides”), which is **IMPORTANT**! Without this, I believe this would lose some functionality of aliases—specifically their ability to inherit `$type` automatically.
  - 99.9% of the time it’s not an issue; this is just coverage for that one weird edge case 😅 

### Resolver Spec problem

Though this proposal does NOT relate to the upcoming Resolver Spec proposal, I’ll share partial syntax that posed a problem:

```json
{
  "sets": [
    { "values": ["foundation.json"] },
    { "values": ["components/button.json"] }
  ]
}
```

Without going too much into detail, part of the resolver spec relied around declaring token “overrides” in individual files (modes/themes/what-have-you). However, the files would have had to be complete, full, valid DTCG format documents. Further, the references MUST be separate documents, otherwise the overrides wouldn’t work.

I believe this would lead to issues where the Resolver Spec would overreach in requiring what tokens went in what files. The DTCG format would swing drastically from “no opinion on files” to “extremely-detailed opinion about files and relative relationships.” At best, it would be a timesink shuffling tokens around between files and file boundaries. At worst, the Resolver Spec wouldn’t work if you declared your files in some magically-incorrect way (especially if you tried to reuse the same token files and apply them in different orders for different themes). And also, within the files, you have no way to reuse parts of other files.

In short, it was not only a user problem; it was an infrastructure problem having to worry about whether local or remote files would be supported, and for design system teams worrying about how to deploy and/or maintain these files in the perfect way.

Instead, `$ref` is a happy compromise because **it allows complete flexibility in file structure.** Because it allows referencing sub-documents anywhere—locally, remotely, even _partial_ documents—you could simply organize tokens in separate files how you wanted to first (which is how all DTCG consumers operate today—being in full control of their file separation), while still being able to opt in to the Resolver Spec. So **this provides a clear upgrade path with more flexibility,** without asking anything in return.


c1rrus:
I've literally _just_ read this, so this is a bit of a knee-jerk reaction (perhaps I'll feel differently after sleeping on it :-P), _but_... I have concerns about this proposal:

* While I take your point about the prior art and widespread usage and support of JSON Schema, I'd argue that within the domain of design systems and design tokens, the current `{foo.bar.baz}` syntax is more familiar. Mainly because it's inspired by Style Dictionary's syntax, and Style Dictionary has been the most widely used design token export tool for a while now. Newer ones like Cobalt UI/Terrazo support DTCG and therefore also use the current syntax.
* Yes, the current syntax requires a little bit of string parsing to detect a reference versus an actual value, but I'm not convinced that `if ( token.$value.charAt(0) === '{' )` is any more difficult than `if ( typeof token.$value === 'object' && token.$value.$ref !== undefined )`.

Re-reading #166 and thinking about the resolver spec, I wonder whether "refering tokens in another file" is the right problem to be solving. Clearly, there's a strong demand for being able to split up your design tokens across multiple files - both to keep things manageable when you have a large number of design tokens, and to do stuff like theming by varying which files get included by some mechanism. But that doesn't necesarily require references that point to specific files.

In many cases, the intent is to merge all the source files into a single collection of tokens, and _then_ to do stuff with it. This is how tools like Theo and Style Dictionary have been doing it forever and I'd argue the Resolver Spec is another (more advanced) take on that concept too. Resolving design token references doesn't happen until after that file merging process has been done, so there's no need for them to specify a file that a token is in.

We can debate whether or not that kind of merge-and-then-resolve-references approach is the "best" way to go, but my point is enabling tokens spread across multiple files can be achieved without changing the current reference syntax.

So, considering how big of a breaking change this would be, I wonder if it's worth it.

drwpow:
@c1rrus great thoughts, thank you! 🙏 I wanted to think on the points you raised for a few days.

> In many cases, the intent is to merge all the source files into a single collection of tokens, and then to do stuff with it. This is how tools like Theo and Style Dictionary have been doing it forever and I'd argue the Resolver Spec is another (more advanced) take on that concept too. Resolving design token references doesn't happen until after that file merging process has been done, so there's no need for them to specify a file that a token is in.

This is a great callout, and I think really gets at the mental model of how people think about the DTCG spec now, how people _want_ to think about it, etc. So this is exactly the conversation I wanted to have with this proposal, even if it doesn’t move forward.

Specifically “Resolving design token references doesn't happen until after that file merging process has been done” doesn’t have to be how this works! Again, see the comment from the JSON spec:

> Attempting to remove all references and produce a single schema document does not, in all cases, produce a schema with identical behavior to the original form.

I interpret this as working exactly as aliases work today—you don’t have to preresolve those preemptively, and you can (and should) preserve all the pointers as actual pointers (not deep clones of the data). I do realize the spec changes as-written are confusing about that, so I should update the examples. But this can work the exact same as aliases do. And leaning on how JSON pointers are used today, the filenames themselves are negligible, again, they can work the same. It just removes the restriction of keeping tokens in files _by necessity_ and moves it to a pattern where you can put any token in any file _by choice_. You can point to tokens in the same file, or remote files. Up to you.

> Yes, the current syntax requires a little bit of string parsing to detect a reference versus an actual value, but I'm not convinced that if ( `token.$value.charAt(0) === '{'` ) is any more difficult than if ( `typeof token.$value === 'object' && token.$value.$ref !== undefined` ).

Yeah performance wasn’t a primary motivator of the change, but this is mostly only true in JavaScript. I agree with you—it is fairly-trivial to parse and typecast on-the-fly. But in other languages that rely more heavily on fixed structures and memory allocation, leaning on typecasting that’s apparent at the schema level (like [serde](https://serde.rs/) for Rust) is much simpler. Again, not to mention the fact that for JSON `$ref`s, every language already has existing tooling to understand, parse, and resolve these.

---

As an aside, after writing this I did think about the relation between this and the Resolver Spec more, and I reached the conclusion the Resolver Spec in its current form is sufficient, and whether we accept or reject this proposal won’t change that. The problems I also _thought_ I would hit in the Resolver Spec ended up not being problems at all when I worked through a [complex example](https://gist.github.com/drwpow/6e9902f2815f27f7ae8f87370bff5b33)!

So all that to say I’d like to evaluate this solely on the basis of:

1. Does this solve problems for the current state of DTCG, and DTCG alone? (I think so)
2. Does this “stand on the shoulders of giants” and use familiar, prior art? (I think so)

And we may come to the conclusion “no” and that’s OK with me. I’m happy whether this is accepted or rejected; so long as we evaluated it.

markacianfrani:
As a tool author, it is very difficult to reliably tell if a given spec is a single file vs. multi file with remote refs. I don't work with JSON Schema enough but there's enough prior art that I could find easily find a solution within 5 minutes so this looks pretty good. 

At the same time, parsing "{" also wasn't that big of a pain point for me. And it does push us farther from human readable but I think that ship has sailed.

I also would encourage more breaking changes. As a consumer, I understand the (draft) contract.

technical-reports/format/terminology.md
```diff
@@ -104,14 +104,33 @@ Groups are arbitrary and tools SHOULD NOT use them to infer the type or purpose
...
-The following Sass example illustrates this concept:
+The following CSS example illustrates this concept:
 
-```scss
-$color-palette-black: #000000;
-$color-text-primary: $color-palette-black;
+```css
+:root {
+  --color-palette-black: #000000;
+  --color-text-primary: var(--color-palette-black);
```
drwpow, line 112: Note: I also updated this earlier section introducing the term “Alias,” and used CSS variables rather than Sass variables. I think it better preserves this idea we want to keep—of aliases carrying through to runtime better. Plus I think in 2024, CSS variables are more widely-used than Sass variables (completely guessing; I have no data to back this up).

drwpow:
> As a tool author, it is very difficult to reliably tell if a given spec is a single file vs. multi file with remote refs. I don't work with JSON Schema enough but there's enough prior art that I could find easily find a solution within 5 minutes so this looks pretty good.

That’s a great point, and it is a little bit of a papercut for sure—you don’t always know the depth of schemas that refer to other schemas until you try resolving it. But this is also a common problem that has over a decade of prior art and workarounds, too, so it’s not new.

This goes beyond this PR, but in many setups today, people _are_ already using multiple token files for their design system. And they have to manage and describe that “meta” layer that determines how everything fits together. By using JSON pointers (`$ref`s), some people MAY want to just have a single document that describes their entire token structure, and offload maintaining/building that meta layer. This gives them the option to do so.

Of course, there are also more complex usecases with modes, etc., that will be solved by the still-in-progress Resolver proposal (that I have seen a preview, and am a big fan of!). So if we accept JSON pointers, we don’t require people to use multiple documents at all. We just give people better options to organize tokens in a way that works best for them without guiding them toward complexity or a self-contradictory design.

kaelig:
@drwpow walked me through the PR, and I grilled him on various aspects to ensure we weren't missing anything. I feel comfortable with this proposal, as it helps with adoption and tool creation (JSON ref libraries already exist for several languages).

drwpow:
Started playing around with an implementation in Terrazzo, and I realized there’s an error with the proposal that needs to be corrected. And it may affect some peoples’ opinions about the approach:

#### 1. `$value` shortcuts

The way this is described, both would essentially be the same:

```jsonc
{
  "color-blue-500": { "$type": "color", "$value": "#218bff" },
  "color-action-bg": { "$ref": "#/color-blue-500" }
}
```

```jsonc
{
  "color-blue-500": { "$type": "color", "$value": "#218bff" },
  "color-action-bg": { "$type": "color", "$value": { "$ref": "#/color-blue-500/$value" } }
}
```

The difference in the latter is the `$value` MUST be required, otherwise the reference would be invalid (the value for a color token couldn’t be a token itself). But that’s a little bit of extra boilerplate, isn’t it? We could take 2 approaches

##### Solution 1: Adjust the spec (simple, correct, but verbose)

We could simply require the full value: `"$value": { "$ref": "#/color-blue-500/$value" }`. This is correct. And it’s requiring precision because it’s necessary.

##### Solution 2: Allow reserved words (more complex, more ambiguous, but less verbose)

Or we could just say “when the key its own is a reserved key (`$value`, `$type`, `$description`, etc.), and the reference resolves to have that **same reserved word**, pretend the alias points there, e.g.:

- ✅ `"$value": { "$ref": "#/color-blue-500" }`: `color-blue-500` also has `$value` so it can be shortened
- ✅ `"$value": { "$ref": "#/color-blue-500/$value" }`: still allowed because it’s a full pointer
- ❌ `"$description": { "$ref": "#/color-blue-500" }`: **error** because `color-blue-500` doesn’t have a `$description` key defined

The obvious downside is **this increases ambiguity** and introduces the concept of conditional lookups, which is usually a bad practice. This may also not be how toolmakers implement JSON pointers naïvely, so it could likely lead to errors. But again, it is easier-to-type.

#### 2. Alias boundaries

Not an “error” per se, but an omission that needs to be clarified, is **what, then, counts as an alias**? For example, if we were to say “`color-action-bg` aliases `color-blue-500`,” which of the following code would still keep that statement true?

1. `"color-action-bg": { "$ref": "#/color-blue-500" }` (whole-token)
2. `"color-action-bg": { "$value": { "$ref": "#/color-blue-500/$value" } }` ($value only)
3. `"color-action-bg": { "$description": { "$ref": "#/color-blue-500/$description" } }` ($description, or any other non-$value property)

The difference here is **aliases can’t be “flattened” until the very last step** under the current language, and the reference must be preserved. But if users can compose any part of their schema, what constitutes an “alias” now? 

##### Solutions

How should we amend the language?

1. A token is an alias of another token if its `$value` points to another token’s, e.g.:

    - ✅ `"color-action-bg": { "$ref": "#/color-blue-500" }`: this is an alias because $value is aliased by proxy of the entire object pointing to another token
    - ✅ `"color-action-bg": { "$value": { "$ref": "#/color-blue-500/$value" } }`: this is an alias because $value directly points to another token’s
    - ❌ `"color-action-bg": { "$description": { "$ref": "#/color-blue-500/$description" } }`: this is NOT an alias because $value does not point to another token

2. A token is **only** an alias of another token if the entire token points to another token AND $value is not overridden, e.g.:

    - ✅ `"color-action-bg": { "$ref": "#/color-blue-500", "$description": "Action BG color" }`: this is an alias because the entire token points to another token AND $value is not overridden (even if some other properties are)
    - ❌ `"color-action-bg": { "$ref": "#/color-blue-500", "$value": "#218bff" }`: this is NOT an alias because even though the token originally pointed to another token, $value was overridden making this a unique token.
    - ❌ `"color-action-bg": { "$description": { "$ref": "#/color-blue-500/$description" } }`: this is NOT an alias because $value does not point to another token

3. A token is **only** an alias of another token if $value is aliased directly, e.g.:

    - ❌ `"color-action-bg": { "$ref": "#/color-blue-500" }`: this is NOT an alias because it is happening one level up from $value
    - ✅ `"color-action-bg": { "$value": { "$ref": "#/color-blue-500/$value" } }`: this is an alias because $value directly points to another token’s
    - ❌ `"color-action-bg": { "$description": { "$ref": "#/color-blue-500/$description" } }`: this is NOT an alias because $value does not point to another token

Out of all the approaches, #3 would be beneficial being the **easiest to statically-analyze.** You could tell which tokens are aliases without having to resolving any pointers. But #3 would be the most restrictive, too, **preventing people from aliasing entire groups** like other methods could.

As the spec proposer, just for the sake of argument, I would probably stick with #1 where even though it’s hard to statically-analyze the number of final tokens without doing work resolving everything, that seems like a worthwhile tradeoff to give schema authors more raw power to generate tokens and create aliases more freely without restrictions. In other words, I don’t really know if it’s advantageous to make token counts easier up front, especially considering the aliases have to be resolved one way or another, and all approaches are the same amount of work in the end.

---

I’ll update the PR description with **Solution 1** and **Solution 1** respectively as my rough proposals, but I could be easily swayed 🙂. Would really just love thoughts in general and poking holes in this more.

ilikescience:
**tl;dr:** i don't think we should replace the current alias syntax with the `$ref` syntax:
1.  i like the benefits of introducing pointer syntax to allow for more precise aliases and overrides, but i think these can be accomplished in a syntactically compatible way with the current (object notation) format. 
2. making the jump to multi-file data is complicated and i think we should consider it separately from syntax/format

---

Long version / stream of consciousness:

There's one few minor question marks and one major question mark in my head:

1. Minor: going away from {foo.bar.baz} does lose the nice hint that tokens are meant to be nested as javascript objects. `#/color/blue/500/$value` feels less javascript-object-y than color.blue.500.$value, and carries the connotation (to me at least) that these are folder structures instead. However, in my adventures writing tools, I tend to just 'flatten' the dtcg JSON to make it easier to access tokens, so i don't see one being performance-wise better than the other.

2. Minor: As far as I can tell `$ref` was originally designed for JSON schema, not JSON data itself. While that alone isn't disqualifying, it does make me suspicious that it might not be the right tool for the job; there are a lot of consequences to adopting it wholesale that we will have to pick through.

3. Very minor: in my opinion escaping `/` and `#` with `~0` and `~1` is pretty esoteric, and makes my eyeballs itch a little.

4. Major: a token file might be syntactically valid (ie contain refs formatted correctly), but aliases may not be possible to resolve — if the file you're referencing has moved, or if it's on a server that can't be accessed. I really like the implied feature of the current format, which is that a file with nesting and aliases can be statically analyzed for circular references, and if none exist, it can be safely[^1] converted into a flat dictionary with no aliases. I feel like making the jump to multi-file should be separated out from the other ideas, which can be brought through under single file constraints.

For example, I agree that being able to grab more than just the `$value` of an aliased token is valuable. There a way to refine the current object notation in a way that allows for things like:

```json
{
  "color":{
    "blue":{
      "$type":"color",
      "$description":"The color blue",
      "$value":{
        
      }
    },
    "brand":{
      "$type":"color",
      "$description":"our brand color",
      "$value":"{color.blue.$value}"
    }
  }
}
```
Overrides are tricky, but I do think that we could add that to the current spec without breaking backwards (syntactic) compatibility. 

```json
{
  "color.brand": {
    "$value": "{color.blue}",
    "$description": "our brand color"
  }
}
```
To a few of your points:

> When a schema isn’t distinguishable by structure, tool makers have to do additional work string matching to do basic typecasting and discrimination.

and

> Out of all the approaches, https://github.com/design-tokens/community-group/pull/3 would be beneficial being the easiest to statically-analyze. You could tell which tokens are aliases without having to resolving any pointers.

I agree that trying to differentiate an alias from a non-alias string feels somehow philosophically _wrong_, but I haven't had trouble detecting aliases with the current spec. I'm certainly not the most experienced implementer, so my attempt might be naive, but in my resolver I have this function:

```ts
function isTokenAlias(tokenValue: TokenValue): boolean {
  return typeof tokenValue === "string" && tokenValue.startsWith("{") && tokenValue.endsWith("}");
}
```

Or you could use the regex `/^\{.*\}$/` if you like that kind of thing.

Identifying an alias by looking at the key and asking if it is === `$ref` seems philosophically more correct, in in practice takes the same number of steps and is still, ultimately, string matching.

Either way, I'm for the spirit of proposal #1, in terms of allowing someone to alias an entire object or to cherry-pick attributes.

> Attempting to remove all references and produce a single schema document does not, in all cases, produce a schema with identical behavior to the original form.

I take this to mean that in JSON schema world, referencing another schema or part of a schema is both semantically and functionally different than "flattening" the reference into a single static document. For example, if our schema for dtcg files references a specific json schema, we're saying "if that json schema is updated at some point, the updates apply to our schema, too." This is different than simply copy-pasting the schema into ours, which loses the dynamic relationship between the two.

I believe the current alias format (object notation) respects both the semantic and functional differences between "foo is an alias for bar which has the value of baz" and "foo has the value of baz". 

In a single file world, there is no functional difference at all, but the semantic difference is important and remains.

The only place where I think `$ref` becomes valuable is in a multi-file world. 

[^1]: By "safely", I mean the semantics are preserved, eg a chain of nested references like color.brand.primary → color.blue.500 → #0000ff chain will deterministically be resolved to "color.brand.primary" = "#0000ff".

DarioSoller:
> The only possible issue is that we'd no longer support "inline aliases" interpolated in a value. For example: {spacing.inline.small} 1em 0 {spacing.inline.small}. I don't know if this is a limitation we want to embrace or if there are legitimate use-cases where interpolation is needed. Editors & community, please advise!

@kaelig Losing the possibility for shorthand design token values, takes away a lot of flexibility we currently have to create systematic design system approaches.

This also sounds like it wouldn't be possible anymore to do more algorithmic design system approaches. For example combining two or more other design tokens in a calculation, in order to calculate the value of a token?

Both would be a real bummer IMO.

ilikescience:
> On this one point I have to say I've resented, in the past, the fact that `.value` gets preferential treatment and that I can't reference an entire subpath of my token tree, or a specific field like a description or attribute. There are relations in my tokens that would allow me to write DRYer code, that I feel I can't express with aliasing being reserved to values.

@Sidnioulz check out the proposal I put together in https://github.com/design-tokens/community-group/pull/298 — it brings in a lot of @drwpow's suggestions from this one, but without breaking changes. To wit, it opens up the possibility of referencing subpaths!

brettdorrans:
Hi all, I've discovered this specification in my exploration of building a unified design token linter/code generation tool (apologies, I've embarked on this before discovering the prior art!) I hope you don't mind me dropping in with some comments, from an outsider perspective.

Reading through this, and the wider spec as it currently exists, I see lots of things borrowed from JSON Schema. It seems a lot of effort is going into discussion of schema internals e.g. multi-file references, handling edge cases. These are totally valid points, but they make me question: why are we inventing our own bespoke referencing and validation mechanism at all?

There are established, well-designed specifications for describing and linking JSON documents - JSON Schema being one we appear to be re-implementing indirectly here. They already define things like `$ref`, syntax, remote references etc. Rather than debating precise schema implementation details like whether to escape `/`, could we not simply choose an existing superset and adopt it wholesale? We can then focus energy on the aspects of this spec that are domain-specific i.e. design tokens.

In practise, this might look like:

- **Picking a base spec**: e.g. we might decide a design token file is a JSON document validated by JSON Schema. We might decide it's JSON Pointer syntax. The point is not which one we choose, but that we don't design our own.
- **Reference, not replicate**: we should avoid re-implementing existing specs, or straying outside our domain (design token interchange format).

I think we should decide how coupled we want the specification to be to existing tooling. We should avoid artificially limiting the robustness of this specification because migration may be difficult for downstream consumers. There are no legitimate downstream consumers yet, this specification is not final. There is still time to make a clean break.

My concern is that we seem to be spending time debating custom syntax and validation when those concerns are orthogonal to design tokens themselves. The scope of this group is to define a **design token interchange format**, not to create a new flavour of JSON. Using an existing JSON superset would let us stand on existing tooling and remain focused on our raison d'etre.

Thanks again for the work you're doing here, I am here also because I've encountered the pain this group is trying to solve. I hope you take my drive by commentary in good faith!

brettdorrans:
> @brettdorrans Thanks for your feedback! I think we are aiming at the same things overall.
> 
> > There are established, well-designed specifications for describing and linking JSON documents - JSON Schema being one we appear to be re-implementing indirectly here. They already define things like $ref, syntax, remote references etc.
> 
> This is a proposal to align DTCG spec around exactly what you specified—existing standards and prior art. This quite literally adopting part of JSON Schema. Would you say you are in favor of this proposal to the spec? Or are you saying “we should have started here in the first place?”

Hey @drwpow - thank you for your efforts on this project. Apologies if my comments were unclear. I'm glad to hear the goal is to align with existing standards, and I appreciate this PR moves further towards that goal.

I do support adopting JSON Schema's syntax and pointer semantics as part of the spec, but I think we should go further. Right now this PR focuses on introducing `$ref` for aliases. My view is that the design token format shouldn't reinvent JSON-Schema-kike features piecemeal; instead a better approach may be:

- Adopt JSON Schema across the entire spec, not adopt parts of JSON Schema or reimplement its pointer system. 
- Use `$ref` universally, not just for `$value` - if we adopt JSON Schema holistically, we can allow tokens to reference arbitrary parts of other tokens (or documents) in future. We open up a lot more interoperability this way. We shouldn't need to define what an alias is or how `$ref` works - we should point to JSON Schema docs.
- Avoid re-implementing validation logic or enforcing this too tightly in the spec - JSON Schema tooling already handles remote references, recursion, circular‑reference detection, etc. If we fully embrace it, tool authors can use existing libraries instead of writing custom resolvers for DTCG‑specific syntax (this is the worst case scenario imo!)

I appreciate that this PR is trying to align DTCG with prior art. From my perspective it's a good start, but it doesn't yet resolve the bigger question of why we are debating syntax for a referencing/validation mechanism at all.

Leveraging JSON Schema and JSON Pointer wholesale would allow the specification to focus on design‑token concepts rather than on schema‑design edge cases, and it would future‑proof the format by keeping it compatible with a broad ecosystem of tooling.

lukasoppermann:
>The only possible issue is that we'd no longer support "inline aliases" interpolated in a value. For example: {spacing.inline.small} 1em 0 {spacing.inline.small}. I don't know if this is a limitation we want to embrace or if there are legitimate use-cases where interpolation is needed. Editors & community, please advise!

@kaelig 

If I understand this correctly while the parts a separate like in typography tokens every part could have it own reference, right?

Wondering if this creates an issue for semi-transformed tokens, like we have in style-dictionary @jorenbroekema?

brettdorrans:
> > This is true. The spec is not final, so any current downstream consumers are using it at their own risk. Breaking changes are to be expected, and we cannot (should not!) let the existence of early adopters hobble development of the spec.
> 
> The people who adopted DTCG are the people who are contributing here and who are making it possible to have feedback and understand the edge cases that the format will need to support in v1. It's a long-winded journey.
> 
> Whilst indeed the spec is not final and breaking changes may happen, to say that the people who are using it, providing feedback, and contributing here are not "legitimate" users feels dismissive of their own efforts.
> 

@Sidnioulz Thank you for taking the time to respond - I want to sincerely apologise if my choice of words came off as dismissive. That was most definitely not my intent. I have the utmost respect for the work everyone here is putting into this. 

On reflection, I think I have probably come to this group with the wrong attitude - I am here because I needed a standard, then I found that standard limiting in frustrating ways - so perhaps that frustration clouded my words. I apologise for expressing that frustration in a dismissive way.

My concern was not meant to diminish the effort or legitimacy of the adopters or contributors - 'legitimate' was a poorly chosen word. I completely agree that early adopters and feedback is invaluable. I would consider myself in this group.

drwpow:
Thanks all for input! We’ll be rolling this into #298 which embodies many similar concepts, but takes a more holistic approach to backwards-compatibility and the way it affects tokens and groups.

PR #260, opened 2025-01-15
Add Synology to the list of companies represented on the DTCG

gcaaa31928:
Hello 👋 As a frontend architecture leader, I’d like to list Synology as a DTCG member, as we're becoming increasingly involved with the group and actively integrating the format into our frontend architecture.


kaelig:
Thank you for your support to the Design Tokens Community Group!

PR #264, opened 2025-02-07
Fix transition documentation

### LINKED ISSUE #227: Transition token documentation issue
In the specification of the transition token, we are mentioning "the color of the shadow" under timingFunction.
This seem to be a mistake, as the token is not related to colors or shadows.

> timingFunction: **The color of the shadow**. The value of this property MUST be a valid [cubic bézier](https://second-editors-draft.tr.designtokens.org/format/#cubic-bezier) value or a reference to a cubic bézier token.

https://second-editors-draft.tr.designtokens.org/format/#transition

drwpow:
## Changes

- Fixes #227. Fixes typo in 2nd Editors draft without changing the draft itself.
- Updates `package.json` to be in sync with `main` so it can deploy properly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,26 +9,117 @@
       "version": "1.0.0",
       "license": "SEE LICENSE.md",
       "devDependencies": {
-        "husky": "^8.0.1",
-        "lint-staged": "^13.0.1",
-        "live-server": "^1.2.2",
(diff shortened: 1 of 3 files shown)


kaelig:
@drwpow can you re-apply these changes in `main`, now that editors drafts snapshots are stored there?

PR #266, opened 2025-03-27
Color first draft (updated)

ilikescience:
## Formatting and updating changes:
- Updates #147 to use the latest version of the format module
- Brings suggestions from #257 into #147 
  - Adds required colorSpace key/value
  - Adds required components key/value
  - Adds optional alpha key/value 
  - Adds optional hex key/value
- Removes gradient section altogether — gradient tokens are defined in the format module currently, so having a separate additional section would conflict with the current spec
- Keeps the naming section, but converts everything from scss to design tokens format to reduce confusion
- Reduces the size of some of the naming examples which got verbose as dtcg format
- Removes the section from naming conventions about supporting multiple color themes, as it had syntax and format that didn’t match the current spec

## Additional changes to discuss:
- Adds "supported color spaces." We haven't discussed this, but I ended up down a rabbit hole and added them. Essentially, this makes explicit what the allowed values of `colorSpace` should be: all the modes with color functions in CSS Color Module 4, plus Display P3 (supported in FIgma and most platforms) , and XYZ65 (the foundation of all other color spaces, useful as a universal interchange space).
  - I made an editorial decision to omit srgb-linear, a98-rgb, prophoto-rgb, and rec2020 from the list. I'm open to including them for parity with CSS Color Module 4, but worry that supporting these spaces and their different gamuts puts a pretty substantial burden on translation tools — of course with libraries like color.js and culori it's possible to offload this burden, but then you're taking on a dependency.
  
## Things I didn't change, but we should discuss:
- I’m not quite sure what the ‘Example of supporting multiple themes and brands in React’ section is doing, but i left it in for now.
- The ‘aliases’ section in the terminology component of the format module uses scss to discuss aliases, we should consider using dtcg format json instead.


technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+    }
+  }
+}
+```
+
+</aside>
+
+### Alias
+
+A design token’s value can be a _reference_ to another token. The same value can have multiple names or aliases.
```
drwpow, line 39: This is a w3c spec thing—try and use `MAY` or `MUST` (both capitalized) to clarify what is normative vs non-.

```suggestion
A design token’s value MAY be a _reference_ to another token. The same value MAY have multiple names or aliases.
```

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+
+</aside>
+
+## Naming strategies
+
+There are many naming options when it comes to design tokens, especially color type tokens. We’ve identified two that seem to be most commonly implemented, **descriptive and numerical**.
+
+### Base Tokens
+
+For **Base tokens**, here’s how they may be represented in each version:
```
drwpow, line 97: etc etc

```suggestion
For **Base tokens**, here’s how they MAY be represented in each version:
```

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+      }
+      // etc...
+    }
+  }
+}
+```
+
+</aside>
+
+**Bounded scales**
```
drwpow, line 219:
```suggestion
##### Bounded scales
```

drwpow (commented):
This is looking really good! Main highlight is just syncing this up with CSS Color Module 4 wherever possible in terms of specific number ranges and min/max values etc.

Re: supported colorspaces we’re open to supporting a subset of Color Module 4 if there are really clear, defined reasons! But as some of that spec’s editors highly encouraged us to match parity with that, and are expecting that, I’d like to just have “match Color Module 4” as the starting point and for every deviation we should make it clear why we can’t implement X, Y, Z, etc.

caoimghgin:
I'm late to the game, but let us consider that parsing color values in the generation of the design-token.json may not be the proper responsibility for design-token JSON files. Simply put, the spec does not need to declare a key/value for colorspace, components, or alpha to be of use.

Because the CSS Color() spec is well defined and requires a few simple regex to properly lift the values and the color space/model, I suggest the transformer (Style Dictionary) is responsible for parsing any valid CSS Color() string given within the $value key. So long as the color is valid, it can be parsed into any other space/profile as desired for the target platform with no visual difference (except out-of-gamut, of course). 

Still, a fallback color space/model could be valuable where we do not want (or trust) the transformer to convert into another color space (reasons). If so, consider reuse of the CSS pattern for font-family, where we comma delim the values `font-family: "Times New Roman", Times, serif;`. 

A proposed example...

```
{
    "Midtone gray": {
      "$type": "color",
      "$value": "#6a6a6a"
    },
    "Magestic magenta": {
      "$type": "color",
      "$value": "hsl(329, 52%, 52%)"
    },
    "Simple sage": {
      "$type": "color",
      "$value": "hsl(104, 39%, 74%, 0.75), #B1D7A3"
    },
    "Barbie pink": {
      "$type": "color",
      "$value": "oklab(53.76% 0.2324 -0.02)"
    },
    "Overlay gray": {
      "$type": "color",
      "$value": "oklch(44.95% 0 0 / 33.33%), #55555555"
    },
}
```

The advantage is leveraging known patterns/standards established by CSS, maintaining separation of responsibility by allowing transformers to do what they are built to do, and simplifying the output in JSON. 

I'd further propose the choice of output space/model be embedded in Style Dictionary according to design/business decisions per platform and/or brand. All colors could be transformed to hex for pure CSS output, but SCSS may respect the first definition in comma delim array, (hex, hsl, oklab, oklch). For iOS, everything could be transformed to P3 or kept in Hex, or even a mix if possible.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
-# The color type
+# Color type
 
-Represents a 24bit RGB or 24+8bit RGBA color in the sRGB color space. The `$type` property MUST be set to the string `color`. The value MUST be a string containing a hex triplet/quartet including the preceding `#` character. To support other color spaces, such as HSL, translation tools SHOULD convert color tokens to the equivalent value as needed.
+## Format
 
-For example, initially the color tokens MAY be defined as such:
+For color tokens, the `$type` property MUST be set to the string `color`.
```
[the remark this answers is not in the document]
ilikescience: I'll add that!

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+}
+```
+
+</aside>
+
+## Supported Color spaces
+
+The following values are supported for the `colorSpace` property. The `components` array will vary depending on the color space.
+
+| Color space | Value         |
```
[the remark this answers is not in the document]
ilikescience: I'm ok with adding them in! 

My primary concern was just keeping the test surface small for implementation; while I see display-p3 and cielab gamuts used in design tools and production apps, i haven't yet seen a design tool (well, excepting photoshop) support linear rgb, a98rgb, prophoto, or rec2020. I did a cursor search of code on github to see if these spaces were used in production and found almost exclusively color libraries' tests.

Totally understand the desire to pin to Color Module 4 to build off their rigorous foundation, though.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+| ----------- | ------------- |
+| sRGB        | `srgb`, `rgb` |
+| HSL         | `hsl`         |
+| HWB         | `hwb`         |
+| LAB         | `lab`         |
```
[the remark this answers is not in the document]
ilikescience: I'll make sure to use CIELAB.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+| LCH         | `lch`         |
+| OKLAB       | `oklab`       |
+| OKLCH       | `oklch`       |
+| Display P3  | `display-p3`  |
+| XYZ         | `xyz`         |
```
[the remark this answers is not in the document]
ilikescience: Makes sense, I'll add those in.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+
+### sRGB
+
+sRGB was the standard color space for all CSS colors before CSS Color Module 4. It is the most widely used color space on the web, and is the default color space for most design tools.
+
+#### Components
+
+`[Red, Green, Blue]`
+
+- Red: A number between `0` and `255` representing the red component of the color.
```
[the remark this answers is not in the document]
ilikescience: That makes sense. I'll use CSS Color Module 4 as the reference here; there are some values that it bounds from 0-100 instead of 0-1 (hsl, hwb, lab, lch).

A semantic note is that I'm intentionally using `number` instead of `float` since json doesn't have a `float` type. JSON `number` is inclusive of integers and floats and can be signed and/or use scientific notation.

(as a side note, I think we should be super-explicit about this terminology in the format module and then use it very precisely throughout the spec)

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+
+```json
+{
+  "Hot pink": {
+    "$type": "color",
+    "$value": {
+      "colorSpace": "srgb",
+      "components": [255, 0, 255],
+      "alpha": 1,
+      "fallback": "#ff00ff"
```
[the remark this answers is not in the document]
ilikescience: Ok, I'll use hex!

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,354 @@
...
+### LAB
+
+cieLAB is a color space that is designed to be perceptually uniform.
+
+#### Components
+
+`[L, A, B]`
+
+- L: A number between `0` and `100` representing the lightness of the color.
+- A: A signed number between representing the green-red axis of the color.
```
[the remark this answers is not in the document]
ilikescience: In the CSS Color Module, you can use either percentages _or_ numbers in the values for lab & oklab A & B.

In the underlying color space, A and B are unbounded; [l: 0.5, a: 200, b: -300] is a valid color.

From the spec:

> The second and third arguments are the distances along the "a" and "b" axes in the Lab color space, as described in the previous section. These values are signed (allow both positive and negative values) and theoretically unbounded (but in practice do not exceed ±160).

clamping to -125 → 125 is a convenience thing, which allows for percentages to be used.

imo allowing percentages or numbers can be very confusing; while in css you can use the % to disambiguate, 50% isn't a valid JSON number. If the value of a component is `0.5`, how do you know if that's a percentage or not?

For that reason, and a few others (I can go into more details about Chroma here) I suggest we use numbers at all times, and where the color module spec has a [0-1] range (like srgb) it will be the same difference.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+        }
+      }
+    }
+  }
+}
+```
+
+</aside>
+
+### Example of supporting multiple themes and brands in React
```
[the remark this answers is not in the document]
ilikescience: ok!

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+        }
+      },
+      "100": {
+        "$type": "color",
+        "$value": {
+          "colorSpace": "srgb",
+          "components": [187, 187, 187]
+        }
+      }
+      // etc...
```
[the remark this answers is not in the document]
ilikescience: i'll switch it to jsonc for anything that has comments — it seems like these error highlights are just a github thing, i don't see any issues when it's rendered.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+}
+```
+
+</aside>
+
+#### Numerical
+
+**Ordered scales**
+
+Numerical tokens often follow a scale to help delineate how the colors progress. For example, when using an ordered scale, `$color-blue-500` may be the base color, where the lightest color value is `$color-blue-100`, and the darkest value could be `$color-blue-900`, and these values are ordered in increments of 100s in between. We recommend not using sequential numbers (ex: 1, 2, 3, 4) for scalability in case future colors need to be added in between two existing colors.
```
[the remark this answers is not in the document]
ilikescience: ya, I'll update that.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
...
+    }
+  }
+}
+```
+
+</aside>
+
+#### Numerical
+
+**Ordered scales**
```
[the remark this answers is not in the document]
ilikescience: Good callout, I'll update those!

technical-reports/color/color-type.md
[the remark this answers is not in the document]
drwpow: Well said! Thanks for taking the time to discuss this.

> I suggest we use numbers at all times, and where the color module spec has a [0-1] range (like srgb) it will be the same difference.

I’m on board with this.

ilikescience:
@drwpow thanks for the extensive review and all your notes! I've made the changes you requested unless otherwise noted. Feel free to check it out and make sure I didn't miss anything! @kaelig i'd love your review, too!

drwpow:
@caoimghgin Thanks for suggesting this! At this point we **are** conforming to CSS Color Module 4, per your suggestion. So everything you’d like to accomplish will work with our current version of the proposed syntax! But at this time we’re not open to considering “just accept any CSS string” since those are famously NOT easy to parse, and it is NOT a simple regex to get the values (they support arbitrary whitespace, numbers and percentages, in many functions are very lax when it comes to commas and other characters, etc.). It’s only trivially parseable if you defined a very limited subset of what CSS allows.

But the good news is we’ve done that! We have a nearly 1:1 translation of CSS Color Module 4, just in JSON object notation form. No RegEx needed, and static JSON is much friendlier as a universal interface to every programming language. JS may make parsing, data types, and memory assignment easy, but many other languages do not and parsing CSS would be a significant technical barrier and beyond the scope of this format.

technical-reports/color/color-type.md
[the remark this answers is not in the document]
drwpow: Cool. Yeah don’t feel strongly about `float`. @kaelig may have thoughts on exact terminology (I think we use `integer` and `floating-point` in places but can’t say it’s 100% consistent)

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
+For color tokens, the `$type` property MUST be set to the string `color`.
+
+The `$value` property can then be used to specify the details of the color, The `$value` object contains the following properties:
+
+- `colorSpace` (required): A string that specifices the color space. For supported color spaces, see the [supported color spaces](#supported-color-spaces) section below.
+- `components` (required): An array representing the color components. The number of components depends on the color space. Each element of the array MUST be either:
+  - A number
+  - The 'none' keyword
+- `alpha` (optional): A number that represents the alpha value of the color. This value is between `0` and `1`, where `0` is fully transparent and `1` is fully opaque.
+- `fallback` (optional): A string that represents a fallback value of the color in [CSS hex color notation](https://www.w3.org/TR/css-color-4/#hex-notation) format. The fallback color MUST omit the alpha value, since alpha is specified in the `alpha` property. The fallback value is used when the color cannot be represented in the specified color space.
```
drwpow, line 16:
```suggestion
- `hex` (optional): A string that represents a fallback value of the color in [CSS hex color notation](https://www.w3.org/TR/css-color-4/#hex-notation) format. The fallback color MUST omit the alpha value, since alpha is specified in the `alpha` property. The fallback value is used when the color cannot be represented in the specified color space.
```

drwpow: Optional suggestion: thanks for capturing the “fallback color MUST omit the alpha value” bit because we were anticipating a confusing scenario where it’s unclear which alpha value would be used if there was a conflict. But perhaps we could make this clearer by saying something like “only hex 6 is supported, not hex 8” in case “omit the alpha value” isn’t clear to some folks.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
 ```json
 {
-  "Majestic magenta": {
-    "$value": "#ff00ff",
-    "$type": "color"
+  "Hot pink": {
+    "$type": "color",
+    "$value": {
+      "colorSpace": "srgb",
+      "components": [255, 0, 255],
```
drwpow, line 26:
```suggestion
      "components": [1, 0, 1],
```

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
   },
   "Translucent shadow": {
-    "$value": "#00000080",
-    "$type": "color"
+    "$type": "color",
+    "$value": {
+      "colorSpace": "srgb",
+      "components": [0, 0, 0],
+      "alpha": 0.5,
+      "hex": "#00000080"
```
drwpow, line 37:
```suggestion
      "hex": "#000000"
```

technical-reports/color/color-type.md
drwpow: sidenote: this is a really clear example! Seeing how `[1, 0, 1]` relates to `#ff00ff` for most folks will make it “click” I think (even if they’re not familiar with CSS Color Module 4)

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
+      "components": ["none", 0, 100],
+      "alpha": 1,
+      "hex": "#ffffff"
+    }
+  },
+  "Black": {
+    "$type": "color",
+    "$value": {
+      "colorSpace": "hsl",
+      "components": ["none", 0, 0],
```
drwpow, line 67: Suggestion: having a `hue: "none"` example above is really good! But maybe an arbitrary color would be good as a second example

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
+
+</aside>
+
+## Supported Color spaces
+
+The following values are supported for the `colorSpace` property. The `components` array will vary depending on the color space.
+
+| Color space  | Value            |
+| ------------ | ---------------- |
+| sRGB         | `srgb`, `rgb`    |
```
drwpow, line 83: Would love @kaelig’s input on this but I know CSS Color Module 4 is really explicit about `srgb`, and not conflating it with RGB which now describes many other colorspaces (Culori does, which IMO shouldn’t be followed). Even as an alias I’m worried it may cause confusion in the future (maybe; open to arguments if it’s helpful!)

```suggestion
| sRGB         | `srgb`    |
```

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
+
+#### Components
+
+`[L, A, B]`
+
+- L: A number between `0` and `100` representing the percentage of lightness of the color.
+- A: A signed number between representing the green-red axis of the color.
+- B: A signed number between representing the blue-yellow axis of the color.
+
+A and B are theoretically unbounded, but in practice don't exceed -160 to 160.
```
drwpow, line 227: 👍

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,503 @@
...
+      "alpha": 1,
+      "hex": "#ff00ff"
+    }
+  }
+}
+```
+
+### XYZ-D50
+
+XYZ-D50 is similar to XYZ-D65, but uses the D50 illuminant.
```
drwpow, line 475: 💯 these descriptions are very good and clear! Wasn’t expecting such an approachable document out of this but it’s a nice bonus 🙂

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,393 @@
...
+<aside class="example">
+
+```json
+{
+  "color": {
+    "green": {
+      "$type": "color",
+      "$value": {
+        "colorSpace": "srgb",
+        "components": [171, 202, 188]
```
drwpow, line 20: Here, and below, normalize to `0`-`1`

drwpow (approved):
This looks great! There are a few examples which need to be updated but overall this gets my stamp. Would love @kaelig to review as well though

caoimghgin:
> @caoimghgin Thanks for suggesting this! At this point we **are** conforming to CSS Color Module 4, per your suggestion. So everything you’d like to accomplish will work with our current version of the proposed syntax! But at this time we’re not open to considering “just accept any CSS string” since those are famously NOT easy to parse, and it is NOT a simple regex to get the values (they support arbitrary whitespace, numbers and percentages, in many functions are very lax when it comes to commas and other characters, etc.). It’s only trivially parseable if you defined a very limited subset of what CSS allows.
> 
> But the good news is we’ve done that! We have a nearly 1:1 translation of CSS Color Module 4, just in JSON object notation form. No RegEx needed, and static JSON is much friendlier as a universal interface to every programming language. JS may make parsing, data types, and memory assignment easy, but many other languages do not and parsing CSS would be a significant technical barrier and beyond the scope of this format.

Thanks for replying. I have no doubt we are conforming to CSS Color Module 4. Having recently pulled code from https://colorjs.io/ and inserted an isValid function as a PR, I see the challenges of “just accept any Color() CSS string”, but it appears color.js solved to a satisfactory level.  Advantage: No need to create another format using JSON object notation.

I'm not certain I've encountered a case where other languages are unable to RegExp as easily as JS, but I haven't used every language. Indeed, I'm struggling to imagine what language aside from JS would be used.

I suppose the communication is that with JSON object notation format, we've invented another standard and we're placing a lot of responsibility on the JS plugin to parse color data from Figma/Sketch/etc where I suspect that responsibility is better placed in the transformer. 

It may seem difficult to separate what is good practice from personal preference, but happy to have a forum to share.

drwpow:
> I see the challenges of “just accept any Color() CSS string”, but it appears color.js solved to a satisfactory level. Advantage: No need to create another format using JSON object notation.

If you’ve dealt with parsing, you’d know that parsing/lexing requires breaking down a very complex string into an abstract syntax tree (AST), many of which are JSON-serializable or a generic enough format to be universal. Taking a universal, JSON-serializable AST format and obfuscating it back into a [DSL](https://en.wikipedia.org/wiki/Domain-specific_language) has no advantages and only downsides. Also, in regards to “no need to create another format,” that’s the entire purpose of the w3c design tokens spec—to outline and create another format. That is the overarching goal of this project. We are defining that format here.

Further, culor.js has NOT solved it to an adequate level. It is still alpha software, does not support the spec fully yet, and has performance problems. Not to say it can’t improve! But we also do not want to prop up this specification on the back of a single open-source alpha project.

> I'm not certain I've encountered a case where other languages are unable to RegExp as easily as JS, but I haven't used every language. Indeed, I'm struggling to imagine what language aside from JS would be used.

Swift, Kotlin, and Java are three common consumers of the DTCG format, building iOS, Android, and macOS apps. Go and Rust are second-tier languages also within scope, being 2 languages that frequently overlap with JS/TS and web. It is directly the DTCG format’s concern to outline a non-web, non-JS centric universal format, to which CSS would be a significant technical barrier. One example worth looking into is Rust’s [serde](https://serde.rs/), which deals with serialization/deserialization between JSON and data structures in Rust, and can automatically handle memory assignment. Adding CSS parsing to, say, a Rust library’s plate for consuming DTCG is a step backwards. JSON is AST-compatible, universal, and there are myriad utilities in every programming language to deal with it as a universal language. Few other languages apart from JS have an adequate CSS parser… because they don’t need one.

---

Again, appreciate the input, but for this specific proposal CSS strings are out of the question at this stage. Please feel free to give specific feedback on the proposed JSON!

Worth pointing out that this PR wasn’t meant to “slip in” changes quietly to the color proposal. It’s been in the works for a couple years and has gotten input from many talented individuals, even authors on the CSS Color Module 4 spec. But even after this PR merges, it’s not a “feedback closed; everything is finalized” situation. Please just see this as _preparation_ for final input/review and we’ll tag you for more input, within the boundaries of the current structure/format.

technical-reports/color/color-type.md
ilikescience: I think that's helpful! I'll make it clear that 6-digit hex format is the only acceptable value.

technical-reports/color/color-type.md
ilikescience: Could you clarify what you mean by an "arbitrarily saturated color with a hue?" Like [180, 50, 0]? Which would be `#000000`, but may be meaningfully different if used in interpolation.

drwpow: That’d work! Just having a numeric `hue` value purely as an example; color doesn’t matter

technical-reports/color/color-type.md
ilikescience: here i was following the convention in CSS Color Module 4 where you can define an srgb color with either the `srgb()` and the `rgb()` function, where rgb is just an alias of srgb. I think that's why Culori allows either — it lets the user dictate which function should appear when outputting to CSS. By definition the color itself is identical no matter which one you use, so in this case I think it's a semantics/syntactics issue.

It's true that in the `color()` function doesn't include `rgb` — if we want to narrow our implementation surface a bit further than just the broad alignment to CSS Color Module 4, we can say up front that we're specifically aligning to the `color()` function and modern syntax. I'm up for that if y'all think it would help us be more precise!

ilikescience: Actually, a note here that may preclude us to sticking strictly to the `color()` function; hsl, hwb, lab, lch, oklab, and oklch aren't supported there, since they're all transforms of other spaces. While in practice limiting scope to the `color()` function doesn't limit what colors can be expressed, I can see authors wanting to use one of those transformations. So, we gotta figure out a way to make it logical which syntax (named functions vs. `color()` function) we are preferring where.

drwpow: These are all great callouts, and we haven’t dug in! Probably worth a larger discussion for sure.

For the purposes of this PR / just updating everything, happy to let you “flip a coin” and decide something for the sake of proposal. But agreed we need to set boundaries on terminology here because just throwing around a link to that spec isn’t enough.

caoimghgin:
Interesting to see a use case where DTCG output from Figma/Sketch would be used 'raw' inside native apps rather than first transformed from generic JSON to platform-specific code. From that POV, a desire to pre-parse colors into JSON key/values and attach a profile/space seems reasonable. However...

1. Parsing/transforming CSS Color() is as easy as parsing a hex string. RegExp is only one tool to make that happen, among other methods of cleaning/parsing. 
2. Transforms should be performed by transform tools, not Obj-C, Swift, Kotlin, Java, Python, at the application layer. Good practice to not re-write business logic differently across different stacks. 
3. It is systemically impossible to quality control transforms when distributed to multiple systems, often hidden from experts and stake holders.
4. Not all languages have powerful color libraries available and often coded differently. Which can lead to different/undesirable results. 
5. Certainly, DTCG is a format, however this proposal is different as **DTCG has never invented a new format for a value**. Instead, pre-existing and well-known formats and units lived in **$value** key such as `{$type: color, $value: #555555}`. I point out that `{$type: color, $value: lab(36%, 0, 0)}`, `{$type: color, $value: rgb(85, 85, 85)}`, {$type: color, $value: oklch(0.45 0 0)}, is an existing format and every bit as parsible as the hex we use every day. Nothing platform specific about it, though it is limited to the horrible sRGB color space which is why we want improve it with this effort.

If using the existing standard notation of CSS Color() becomes a possibility, I'd be thrilled to help. Given there have been years of conversations it seems this ship has sailed so I won't muddy the waters any further.

ilikescience:
@drwpow thanks for the latest round of comments! Should be ready for another review (sorry to keep missing things!). I ended up going a bit wild and formatting the allowed color spaces and values into a table because I got tired of jumping back and forth. (note to CSS Color Module 4 editors, feel free to copy my homework 😛 )

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,768 @@
...
+Represents a color.
 
-For example, initially the color tokens MAY be defined as such:
+## Format
+
+For color tokens, the `$type` property MUST be set to the string `color`.
+
+The `$value` property can then be used to specify the details of the color, The `$value` object contains the following properties:
+
+- `colorSpace` (required): A string that specifices the color space. For supported color spaces, see the [supported color spaces](#supported-color-spaces) section below.
```
kaelig, line 11: Would there be value in defaulting to srgb and make `colorSpace` optional?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,768 @@
...
+  }
+}
+```
+
+</aside>
+
+### The `none` keyword
+
+The `none` keyword MAY be used in the `components` array to indicate that a component is not applicable or not specified. This is useful for colors that do not require all components to be specified.
+For example, in the HSL color space, the `none` keyword MAY be used to indicate that there is no angle value for the color, which will be interpretted differently from a color with a hue angle of 0.
```
kaelig, line 48: We need to explain the concept of missing colors and reference the inspiration behind it: https://www.w3.org/TR/css-color-4/#missing

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,768 @@
...
+            <td>L</td>
+            <td>[0 - 100]</td>
+        </tr>
+        <tr>
+            <td>C</td>
+            <td>[0 - Infinity]**</td>
+        </tr>
+        <tr>
+            <td>Hue</td>
+            <td>[0 - 360)</td>
```
kaelig, line 191: What does the parenthesis mean here?

technical-reports/format/types.md
```diff
@@ -12,42 +12,7 @@ If an explicit type is set, but the value does not match the expected syntax the
...
-$majestic-magenta: #ff00ff;
-$translucent-shadow: #00000080;
-
-// colors-hsl.scss
-$majestic-magenta: hsl(300, 100%, 50%);
-$translucent-shadow: hsla(300, 100%, 50%, 0.5);
-```
-
-</aside>
+Represents a color in the UI. For details on how to represent colors, see the [Color](../color) module.
```
kaelig, line 15: 👍

technical-reports/color/color-type.md
ilikescience: [] is inclusive, () is exclusive. ie, if a range is (0-1), everything _between_ 0 and 1 is in the range, but 0 and 1 are not. 

In the case of polar coordinates, 0 and 360 degrees are equivalent. So, the minimum range to represent every possible hue angle is [0-360). If we allow 360, transformers will be expected to re-normalize the range; so I figure we can eliminate some complexity by excluding 360.

The CSS Color Module 4 spec says this:

> This number is normalized to the range [0,360)[^1]

Do you think we need to explain this in our spec?

[^1]: https://www.w3.org/TR/css-color-4/#hue-syntax

technical-reports/color/color-type.md
ilikescience: Spent some time thinking about this, and my suggestion is to keep it as required. Here's some pros and cons of making it optional and defaulting:

## Pro
- Makes it faster to hand-author tokens
- Would likely make most token files smaller — sRGB is likely to be the most common color space used

## Cons
- Transformers/consumers will need to fill in this info regardless, so it's really not saving any time on parse/transform side

I'm not sure of any other place in the spec where we have an optional/default, either. IMO explicit > implicit.

What do you think?

ilikescience:
@kaelig thanks for your comments! I went through and made updates based on your input.

@resource11 I also added a bunch of updates based on our conversation. Please let me know what you think!

technical-reports/color/color-type.md
ilikescience: I went ahead and added an explanation to the top of the table :)

lukasoppermann:
Hey @ilikescience,

it is not shown in the examples, but can we add `alpha` when referencing colors? This is extremely important, as this is impossible with the current setup.

Maybe like this

```
"shadow": {
      "$type": "color",
      "$value": {
        "components": "{base.color.black}",
        "alpha": 0.53
      }
    }
```

Also are we sure about using `components` as a name for the color values? To me this sounds unexpected, and I am wondering if this makes tokens hard to read for people who just contribute to tokens occasionally. Couldn't we use something like `values`?

technical-reports/color/index.html
resource11, line 1: Hey @ilikescience, we updated Adekunle's and Ayeha's urls in previous commits, yet decided to make that change spec-wide in a separate commit (per @kaelig). Now that we're updating here, will you change Adekunle's url to https://www.adekunleoduye.com/ and Ayesha's to https://github.com/ayeshakmaz please?

technical-reports/color/index.html
```diff
@@ -20,12 +20,13 @@
             name: 'Ayesha Mazrana (Mazumdar)',
             url: 'https://twitter.com/AyeshaKMaz',
           },
-          { name: 'Kathleen McMahon', url: 'https://twitter.com/resource11' },
+          { name: 'Kathleen McMahon', url: 'https://kathleenmcmahon.dev' },
```
resource11, line 23: Hey @ilikescience, we updated Adekunle's and Ayeha's urls in previous commits, yet decided to make that change spec-wide in a separate commit (per @kaelig). Now that we're updating here, will you change Adekunle's url to https://www.adekunleoduye.com/ and Ayesha's to https://github.com/ayeshakmaz please?

technical-reports/color/color-terminology.md
```diff
@@ -0,0 +1,39 @@
+# Color terminology
+
+This section provides a high-level overview of the terminology used in the specification and how it relates to color science.
```
resource11, line 3: **praise:** this section is very thorough!

**suggestion:** can we change "color science" to  "color science and colorimetry" and link:

- [color science](https://en.wikipedia.org/wiki/Color_science)
- [colorimetry](https://en.wikipedia.org/wiki/Colorimetry)

here's why:
- our community base have a range of experiences
-  [cognitive accessibility](https://www.w3.org/TR/coga-usable/) should be a priority in spec writing.
- providing links support both people who like to glance quickly and those who like to read up on a topic

technical-reports/color/color-terminology.md
```diff
@@ -0,0 +1,39 @@
...
+
+<dfn>Color space<dfn>
+
+A color space is a specific organization of colors, which helps in the reproduction of color in both physical and digital realms. It defines a range of colors that can be represented and manipulated.
```
resource11, line 7: **suggestion:** link [color space](https://www.w3.org/TR/css-color-4/#color-space)

technical-reports/color/color-terminology.md
```diff
@@ -0,0 +1,39 @@
...
+
+<dfn>Color model<dfn>
+
+A color model is a mathematical representation of colors within a specific [=color space=]. It defines how colors are represented as numerical values, typically using components.
+
+<dfn data-lt="gamut">Color gamut<dfn>
+
+A color gamut is the complete range of colors that can be represented in a specific color space. It defines the limits of color reproduction for that space.
```
resource11, line 15: **suggestion:** link [color gamut](https://colorjs.io/docs/gamut-mapping#what-is-color-gamut) inline

technical-reports/color/color-terminology.md
```diff
@@ -0,0 +1,39 @@
...
+
+Saturation is the colorfulness of a color relative to its own brightness. It describes how much gray is present in a color. A fully saturated color has no gray, while a desaturated color appears more grayish. It is inherently tied to both [=chroma=] and [=lightness=], especially in models like HSL or HSV. A color can be highly saturated but still appear light or dark depending on its lightness.
+
+<dfn>Lightness<dfn>
+
+Lightness is the perceived brightness of a color. It describes how light or dark a color appears.
+
+<dfn>Chroma<dfn>
+
+Chroma refers to the colorfulness of a color relative to the brightness of a similarly illuminated white. It measures how pure or intense a color appears. In the CIE LCh° color space (Lightness, Chroma, Hue), Chroma is independent of lightness and expresses how far a color is from neutral gray along the chromatic axis.
```
resource11, line 35: **typo:** LCh°. Did you mean LCH?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
-For example, initially the color tokens MAY be defined as such:
+## Format
+
+For color tokens, the `$type` property MUST be set to the string `color`.
+
+The `$value` property can then be used to specify the details of the color, The `$value` object contains the following properties:
+
+- `colorSpace` (required): A string that specifices the [=color space=] or [=color model=]. For supported color spaces, see the [supported color spaces](#supported-color-spaces) section below.
+- `components` (required): An array representing the color [=components=]. The number of components depends on the color space. Each element of the array MUST be either:
+  - A number
```
resource11, line 13: **suggestion:** will you clarify a number a bit more specifically, considering a number can be represented as an integer or a floating point?

for example: "A number (integer or floating point) such as 0, 1.5, or 360.0" reduces ambiguity.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+- `hex` (optional): A string that represents a fallback value of the color. The fallback color MUST be formatted in [6 digit CSS hex color notation](https://www.w3.org/TR/css-color-4/#hex-notation) format to avoid conflicts with the provided alpha value.
 
 <aside class="example">
 
 ```json
 {
-  "Majestic magenta": {
-    "$value": "#ff00ff",
-    "$type": "color"
+  "Hot pink": {
```
resource11, line 22: **nit:** We've standardized to use "acid-green" and "hot pink" in the `groups.md` file, yet here we use only "hot pink" throughout the examples. If we're going to standardize to use one value everywhere but the `groups.md` file, should we choose "acid-green" instead for the `color-type.md` file?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+    "$type": "color",
+    "$value": {
+      "colorSpace": "srgb",
+      "components": [1, 0, 1],
+      "alpha": 1,
```
resource11, line 27: **polish:** considering the `alpha` property is optional, and the [alpha value defaults to 1 under the hood if omitted](https://www.w3.org/TR/css-color-4/#rgb-functions), should we be adding this in every opaque color example? color functions that use

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+}
+```
+
+</aside>
+
+### The `none` keyword
+
+When specifying a color in some color spaces, a value of `0` could be ambiguous. For example, in the HSL color space, colors with a [=hue=] of `0` are red; while a single color like `hsl(0, 0, 50)` would not be rendered as red, it may be treated as a completely desaturated red when interpolated with other colors. Therefore, in certain color spaces, `0` is insufficient to indicate that there is no value for a given component.
+
+[[[css-color-4]]] has introduced the `none` keyword to indicate that a component is missing, or not applicable. For example, in the HSL color space, the `none` keyword may be used to indicate that there is no angle value for the color, which will be interpretted differently from a color with a hue angle of 0.
```
resource11, line 49: **suggestion:** change "interpretted" to "interpolated" and inline link [interpolated differently](https://www.w3.org/TR/css-color-4/#interpolation-missing). We then can remove the sentence on line 51.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+
+</aside>
+
+While both examples will render as white, the first example is more explicit about the fact that the hue is not applicable. This is important when interpolating between colors or mixing colors, as using each color in a can yield different results.
+
+## Supported Color spaces
+
+The following values are supported for the `colorSpace` property. The `components` array will vary depending on the color space.
+
+In this table, brackets `[]` indicate that an extrema are included, parentheses `()` indicate that the extrema are excluded. For example, in the HSL color space, hue is in the range of [0 - 360), which means that `0` MAY be used but `360` MUST NOT be used.
```
resource11, line 101: **polish:** can you add the content in line 101 to an editor's note 
`<aside class="example" title="about the hue syntax">`, link the definition of [extrema](https://en.wikipedia.org/wiki/Maximum_and_minimum) and link [hue is in the range of [0 - 360)](https://www.w3.org/TR/css-color-4/#hue-syntax)?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+        </tr>
+        <tr>
+            <td>B</td>
+            <td>[-Infinity - Infinity]†</td>
+        </tr>
+        <!-- OKLCH -->
+        <tr>
+            <th scope="rowgroup" rowspan="3">OKLCH</th>
+            <td rowspan="3">`"oklch"`</td>
+            <td>L</td>
```
resource11, line 221: **polish:** Lighten instead of L

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+        </tr>
+        <tr>
+            <td>B</td>
+            <td>[-Infinity - Infinity]*</td>
+        </tr>
+        <!-- LCH -->
+        <tr>
+            <th scope="rowgroup" rowspan="3">LCH</th>
+            <td rowspan="3">`"lch"`</td>
+            <td>L</td>
```
resource11, line 191: **polish:** Lighten instead of L, Chroma instead of C, Hue, instead of H

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+        <!-- CIELAB -->
+        <tr>
+            <th scope="rowgroup" rowspan="3">CIELAB</th>
+            <td rowspan="3">`"lab"`</td>
+            <td>L</td>
+            <td>[0 - 100]</td>
+        </tr>
+        <tr>
+            <td>A</td>
+            <td>[-Infinity - Infinity]*</td>
```
resource11, line 181: **polish:** can you hyperlink `*`, `**` and `†` to their respective definitions below?

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+  <span>* In CIELAB, A and B are unbounded but in practice don't exceed -160 to 160</span><br />
+  <span>** In LCH, C is unbounded but in practice doesn't exceed 230</span><br />
+  <span>† In OKLAB, A and B are unbounded but in practice don't exceed -0.5 to 0.5</span><br/>
+  <span>‡ In OKLCH, Chroma is unbounded but in practice doesn't exceed 0.5</span>
+</div>
+
+<details class="note">
+<summary>A note about precision in examples</summary>
+<p>The examples below have varying degrees of precision (i.e. numbers after the decimal). This is done to ensure that the 'fallback' color is exactly the same as the defined color when converted to HEX. In practice, the numbers used to define each component can have any degree of precision.</p>
+</details>
```
resource11, line 334: **polish:** rather than details summary, can you use <aside class="edunote" title="about precision in examples"

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+<details class="note">
+<summary>How does this conform to CSS Color Module 4?</summary>
+<p>To provide a logically consistent approach without creating ambiguity around the format, this spec takes the following approach:</p>
+<ul> 
+<li>If CSS Color Module 4 allows a color space to be referenced by **both** a named function (like `srgb()`) **and** a keyword in the `color()` function, use the format intended for the `color()` function.</li>
+<li>If CSS Color Module 4 only defines a color space **either** as a named function (like `hwb()`) **or** a keyword in the `color()` function (like `rec-2020`), use the syntax indicated.</li>
+<li>If CSS Color Module 4 allows **both** unit values (like `50`) **and** percentages (like `50%`) for a component, use the unit value.</li>
+</ul>
+<p>Using this spec as a reference allows us to focus on the design and implementation of the tokens themselves, rather than the underlying color science.</p>
+</details>
```
resource11, line 345: **polish:** rather than details summary, can you use <aside class="edunote" title="How does this conform to CSS Color Module 4?"

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+
+### HWB
+
+Another [=color model=], a polar transformation of sRGB.
+
+#### Components
+
+`[Hue, Whiteness, Blackness]`
+
+- Hue: A number between `0` (inclusive) and `360` (exclusive) representing the angle of the color on the color wheel.
```
resource11, line 451: **suggestion: ** I... read that sentence 3x before I understood it. Can we try a more frictionless one? How about "A number from 0 to less than 360, representing the angle on the color wheel (where 0 and 360 both indicate red)."

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+
+LCH is a cylindrical representation of CIELAB.
+
+#### Components
+
+`[L, C, Hue]`
+
+- L: A number between `0` and `100` representing the percentage of lightness of the color.
+- C: A number representing the chroma of the color.
+- Hue: A number between `0` (inclusive) and `360` (exclusive) representing the angle of the color on the color wheel.
```
resource11, line 519: **suggestion:** see above :)

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,806 @@
...
+
+OKLCH is a cylindrical [=color model=] of OKLAB.
+
+#### Components
+
+`[L, Chroma, Hue]`
+
+- L: A number between `0` and `1` representing the lightness of the color.
+- Chroma: A number representing the chroma of the color.
+- Hue: A number between `0` (inclusive) and `360` (exclusive) representing the angle of the color on the color wheel.
```
resource11, line 587: **suggestion:** 😄 and here?

technical-reports/color/index.html
ilikescience: will do!

technical-reports/color/color-terminology.md
ilikescience: Generally speaking, LCh°, LCh, and LCH are all interchangeable and are used in various specs and references to refer to the same color model/space. There's historical reasons for hº being the most _technically_ correct version, but I don't think we need to fret too much about LCH being misconstrued as something different from LChº.

I'll pick a convention (all upper) and stick to it throughout the doc. If any CIE color experts are feeling their ears burn, and they want to tell me this is a sin against color science, I'm open to the feedback :)

technical-reports/color/color-type.md
ilikescience: I don't feel that using hot pink here is confusing, but I'm happy to change it to green if you feel strongly. Just know that it's a bunch of work to go back and convert/double check all those references 😅

technical-reports/color/color-type.md
ilikescience: I see your point, but I think there's value in showing each examples as "fully loaded", ie shown with all the optional fields provided. Not a particularly strong opinion, but I think that if we want to omit optional fields from the examples we should also omit hex from the examples.

technical-reports/color/color-terminology.md
resource11: **note:** future opportunity to link to the historical reference of the naming convention after the spec publishes.

technical-reports/color/color-type.md
resource11: Thanks for clarifying, can we call out somehow why we're adding it in f it's still optional and what happens if it's omitted in an editors note?

technical-reports/color/color-type.md
resource11: True. It's still an opportunity to educate. Perhaps an editors note?

technical-reports/color/color-type.md
resource11: Also, is interpreted with one "t"?

technical-reports/color/color-type.md
ilikescience: Ok, I'll update this to a note and add the links. Just a clarification that the range expressions using brackets and parens isn't limited to hue, but for any expression of ranges in the spec.

technical-reports/color/color-type.md
ilikescience: Ok, I'll try!

technical-reports/color/color-type.md
ilikescience: Added a note!

technical-reports/color/color-terminology.md
ilikescience: For this and the other suggestion to link an external resource in the definition, I feel uncomfortable adding links to third-party references here. Because these posts offer much more detail, they open the door to having interpretations and information that conflicts with the spec; not that they're _wrong_, per se, but just that they are not edited to the same standard/to the same purpose. Additionally, these pages might be changed in the future or no longer be available at some point which makes the spec more unstable.

IMO the definitions section of the spec should be very normative, and if we'd like to provide more educational resources we can do so in another non-normative section.

If you feel like these definitions are insufficient for the purpose of the spec, lemme know what's not working and I can edit. Otherwise maybe we can align as a group on how we want to include educational resources consistently throughout the spec?

technical-reports/color/color-terminology.md
ilikescience: See my comment on the other suggestion to link an external resource in the definition!

ilikescience:
@resource11 thanks again for your feedback! I pushed an updated version with most of it addressed, and resolved comments where things have changed in response.

Where there's still open questions/discussion, I've left the comments open.

Please take a look at the latest, add any more edits you have, and let me know what you think about the specific points still open in the comments. Thanks!

technical-reports/color/color-terminology.md
kaelig: It's safe to point to other specs, but I agree that it's not beneficial to link to blogs in normative parts of the spec.

technical-reports/color/color-type.md
kaelig: In this case, the application of `number` varies depending on the color space, and can be granularly defined using JSON Schema.

Two variations come to mind: min/max values, decimal precision.

So it's safe to say "number" here, and we can be more specific further down the spec.

Does that solve the issue?

technical-reports/color/color-type.md
kaelig: I don't mind eitherway, and don't think it's confusing to have hot pink here either.

technical-reports/color/color-terminology.md
resource11: I appreciate both your thoughts on this and I respectfully disagree. For example, take a look at the [normative and informative sections of the CSS Color 4 module.](https://www.w3.org/TR/css-color-4/#normative) Not all urls point to a spec page. Some point to blog posts.

technical-reports/color/color-terminology.md
resource11: See me response!

technical-reports/color/color-type.md
resource11: I think I should mention I'm using [conventional comments](https://conventionalcomments.org/) style in my comments. That means **nit** is non-blocking and user preference.

The education portions above, I feel quite strongly about, though. 😄

technical-reports/color/color-type.md
resource11: Hey there! We agreed on making `colorSpace` required a long time ago, so I'm curious why we're revisiting a topic where the group achieved consensus?

technical-reports/color/color-type.md
resource11: Personally, I look at a spec as an opportunity to add in educational components to reduce ambiguity. In other words, I'm coming from the cognitive accessibility standpoint, @kaelig. For some, the term `number` **only** can raise questions, if that person tends to dig into every detail. While two variations come to your mind, not everyone shares your skill to leap to what that term can encompass broadly. I believe it's important to make that distinction, and it will draw more users that are new to reading a spec.

technical-reports/color/overview.md
```diff
@@ -0,0 +1,17 @@
+## Color Tokens
+
+Color tokens can be used to represent colors in different color spaces. Colors represented in tokens can then be converted to other color spaces by translation tools.
+
+Generally speaking, this module uses [[[css-color-4]]] for reference to concepts and terminology. [[[css-color-4]]] is a W3C Working Draft and is not a final specification. It is subject to change and may not be implemented in all browsers or platforms.
+
+<aside class="note" title="Why CSS Color Module Level 4?">
```
resource11, line 7: can you change to `class="ednote"`? That's the class we use for editor's note green boxes

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,518 @@
+# Token naming
+
+## Categorization
+
+There are 3 main categories of design tokens that we will continue to reference in this specification.
+
+### Base
+
+Base tokens are the lowest level tokens and typically consist of a name and [hexadecimal](https://www.w3.org/TR/css-color-4/#hex-notation) value pair.
```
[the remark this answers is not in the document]
resource11: +1

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,812 @@
...
+    </tbody>
+</table>
+<div>
+  <span><span id="fn-1">*</span> In CIELAB, A and B are unbounded but in practice don't exceed -160 to 160</span><br />
+  <span><span id="fn-2">**</span> In LCH, C is unbounded but in practice doesn't exceed 230</span><br />
+  <span><span id="fn-3">†</span> In OKLAB, A and B are unbounded but in practice don't exceed -0.5 to 0.5</span><br/>
+  <span><span id="fn-4">‡</span> In OKLCH, Chroma is unbounded but in practice doesn't exceed 0.5</span>
+</div>
+
+<aside class="note" title="Precision in examples">
```
resource11, line 335: `class="ednote"`

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,812 @@
...
+<p>The examples below have varying degrees of precision (i.e. numbers after the decimal). This is done to ensure that the 'fallback' color is exactly the same as the defined color when converted to HEX. In practice, the numbers used to define each component can have any degree of precision.</p>
+</aside>
+
+<aside class="note" title="Optional values in examples">
```
resource11, line 339: `class="ednote"`

also, 😍

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,812 @@
...
+<p>The examples below are given with all optional values (alpha, hex) included for the purpose of completness. Defining the alpha property for fully-opaque colors is not required, see [[[#format]]].</p>
+</aside>
+
+<aside class="note" title="How does this conform to CSS Color Module 4?">
```
resource11, line 343: `class="ednote"`

resource11:
Thanks for this solid work and for addressing the final polish, @ilikescience! A few minor things, and we should be good to go.

technical-reports/color/color-terminology.md
ilikescience: In the cases of CSS Color Module 4, I can attempt to infer why some references are to w3c, wcag, csswg, whatwg, etc standards organization, and a few are not — specifically, [Oklab] and [Wolfe] point to things I'd consider a blog post.

Each reference provides an authoritative original primary source for a concept. I don't get the sense that these are provided as educational resources, but as citation.

I think that part of the DTCG's mission should be to educate, but I don't think that the spec is the right place to do it. WAI is a good example of this; they say very explicitly "WCAG is not an introduction to accessibility," but provide lots of educational resources on the WAI site. Could we follow their lead and provide educational resources via designtokens.org instead of via the spec?

technical-reports/color/overview.md
ilikescience: will do!

technical-reports/color/color-terminology.md
resource11: I think this is a **great** topic to bring up to the entire group, @ilikescience, especially since this topic has been discussed often within various meetings over the time we've worked on this spec. It'd be a shame to pivot at the last minute.

technical-reports/color/color-type.md
resource11: Excellent! I'll make note that we will be clarifying this is a fast follow PR. image

resource11 (approved):
This looks great to me, @ilikescience. Thanks for taking on the final details and adding key content! 

Let's take the unresolved comments to the group to align and build consensus. 🚀

technical-reports/color/interpolation.md
TzviPM, line 1: Should we mention something here about the idea of relative colors or color blending introduced in https://drafts.csswg.org/css-color-5? Other platforms such as flutter also support defining colors in a relative way (see https://api.flutter.dev/flutter/dart-ui/Color-class.html) or blending (see https://api.flutter.dev/flutter/dart-ui/Color/lerp.html). I would imagine that this could be useful for other mediums such as print whereby some tokens might be blended differently to adjust for paper type

technical-reports/color/color-type.md
drwpow: > We agreed on making colorSpace required a long time ago, so I'm curious why we're revisiting a topic where the group achieved consensus?

Worth noting that even if consensus had been made within a closed group, no clear, visible record was ever made of that consensus. Or even if it was, we still need to raise it with a wider audience for feedback. If there are relevant notes/thinking from those discussions that aren’t easily-findable please bring them to future discussions so more people are in the loop!

technical-reports/color/color-terminology.md
drwpow-figma: > I think this is a great topic to bring up to the entire group, @ilikescience, especially since this topic has been discussed often within various meetings over the time we've worked on this spec. It'd be a shame to pivot at the last minute.

I’m not weighing in on the final call of how much educational content is present in the final version, but I do feel that this should NOT block this PR from merging. I don’t think simply describing a schema is “pivoting at the last minute,” especially since I don’t see any educational content in the [open PR from 2022](https://github.com/design-tokens/community-group/pull/147) that is still not up-to-date.

Again, I think educational content is a great thing, and I welcome it! But I think it would be a disservice to simply try and shove it into the part of the specification that is only describing schema. I believe we could come up with a more holistic strategy that wouldn’t block this overdue work from landing.

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,393 @@
...
```
kaelig, line 9: This references a hex value pair, but the example below doesn't match that description.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,811 @@
...
 }
 ```
 
 </aside>
+
+For more information on the XYZ-D50 color space, see [[[COLORIMETRY]]].
+
+## Future color space support
+
+Future versions of this spec MAY add support for additional color spaces, depending on adoption and support in design tools.
```
kaelig, line 811:
```suggestion
Future versions of this spec may add support for additional color spaces, depending on adoption and support in design tools.
```

(This is similar to a non-normative paragraph)

kaelig (changes requested):

technical-reports/color/overview.md
```diff
@@ -0,0 +1,17 @@
...
+
+Generally speaking, this module uses [[[css-color-4]]] for reference to concepts and terminology. [[[css-color-4]]] is a W3C Working Draft and is not a final specification. It is subject to change and may not be implemented in all browsers or platforms.
+
+<aside class="ednote" title="Why CSS Color Module Level 4?">
+<p>Color is a complex topic. [[[css-color-4]]] provides a comprehensive baseline:</p>
+<ul> 
+<li>It provides a definition of common color spaces, including gamuts and component coordinates</li>
+<li>It gives technical specifications for translating colors between color spaces</li>
+<li>Its authors are experts in the field of color science</li>
+<li>as a spec it must reach an extremely high standard of implementability and precision</li>
```
ayeshakmaz, line 13:
```suggestion
<li>As a spec it must reach an extremely high standard of implementability and precision</li>
```

ayeshakmaz (approved):

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,393 @@
...
+}
+```
+
+</aside>
+
+Numerical token names can also allow for further specificity when needed. For example, when creating neutral palette tokens (like grays), the scale may increase by increments of 25 instead of 100 at the lightest values, and then increment by 100 thereafter.
+
+<aside class="example">
+
+```jsonc
```
adekunleoduye, line 179: Should we be consistent with only using `jsonc` or `json`?

technical-reports/color/token-naming.md
```diff
@@ -0,0 +1,393 @@
...
+| Easy mapping between different tokens for color contrast. For example, all 100-400 tokens pair with 500-900 tokens in order to create accessible color combinations. | Less memorable and less obvious difference between tokens |
+| If using bounded scales, the token name can help indicate something about the underlying value                                                                       |                                                           |
+
+### Alias Tokens
+
+For **Alias Tokens**, we recommend grouping tokens with similar intentions by prioritizing the category + property within the name. For example, all background color Alias tokens would likely start with `color.background.XXX`.
+
+We recommend avoiding abbreviations. For example, use “background” instead of “bg”, to help with clarity.
+
+Here’s how alias tokens MAY be represented:
```
adekunleoduye, line 311:
```suggestion
Here’s how alias tokens may be represented:
```

adekunleoduye (commented):
Left a couple of comments but this looks good overall!

technical-reports/color/token-naming.md
ilikescience: good catch! I'll fix.

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,811 @@
...
+}
+```
+
+</aside>
+
+### The `none` keyword
+
+When specifying a color in some color spaces, a value of `0` could be ambiguous. For example, in the HSL color space, colors with a [=hue=] of `0` are red; while a single color like `hsl(0, 0, 50)` would not be rendered as red, it may be treated as a completely desaturated red when interpolated with other colors. Therefore, in certain color spaces, `0` is insufficient to indicate that there is no value for a given component.
+
+[[[css-color-4]]] has introduced the `none` keyword to indicate that a component is missing, or not applicable. For example, in the HSL color space, the `none` keyword may be used to indicate that there is no angle value for the color; a color with a hue value of `none` MAY be rendered differently from a color with a hue angle of 0, or result in different colors when [interpolating](https://www.w3.org/TR/css-color-4/#interpolation-missing).
```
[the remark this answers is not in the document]
ilikescience: I'll add the backticks!

technical-reports/color/color-type.md
```diff
@@ -1,22 +1,811 @@
...
+      "alpha": 1,
+      "hex": "#ffffff"
+    }
+  }
+}
+```
+
+</aside>
+
+While both examples will render as white, the first example is more explicit about the fact that the hue is not applicable. This is important when interpolating between colors or mixing colors, as using each color in a can yield different results.
```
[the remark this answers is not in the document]
ilikescience: Good catch — I'll edit and clarify a bit.

technical-reports/color/token-naming.md
ilikescience: I only used jsonc where the examples had comments; technically we could switch everything to jsonc with no meaningful difference, but imo it's probably better to stick to jsonc as the exception.

technical-reports/color/interpolation.md
TzviPM: Done :) https://github.com/design-tokens/community-group/issues/268

kaelig (approved):

kaelig:
Thank you @ilikescience for this amazing contribution in tandem with @resource11, and @adekunleoduye @ayeshakmaz for getting the ball rolling early on 🙏🚀

PR #272, opened 2025-05-13
chore: add a versions overview to README, prevED prop for living draft

jorenbroekema:
See slack for versioning discussion. 

Even if we don't have semver (because it's not really the way of doing things for W3C community group specs it seems), it would help if the living draft that's published on tr.designtokens.org had a backlink to the previous editor draft. 

Furthermore, adding a versions overview at least to the README of this repo is useful for tools to see which versions there are, so they can more easily communicate to their users which version of the spec the tool is compliant with.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,18 @@ As vendors adopt the specification and new requirements appear, the community gr
 - [zeroheight](https://www.zeroheight.com)
 - ex-contributors to [Theo](https://github.com/salesforce-ux/theo) (by Salesforce)
 
+## Versions
+
+Here's an overview of all of the published versions of this draft specification:
+
+| name            | url                                      | date       |
+| --------------- | ---------------------------------------- | ---------- |
+| editors-draft/1 | first-editors-draft.tr.designtokens.org  | 2021-09-23 |
+| editors-draft/2 | second-editors-draft.tr.designtokens.org | 2022-06-14 |
+| living-draft    | tr.designtokens.org                      | 2025-04-18 |
+
+> Note: tools can use the date as a version number to signify compliance. For example: `20250418`.
+
 ## Contributing
 
 See [CONTRIBUTING.md](https://github.com/design-tokens/community-group/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -90,6 +90,18 @@ As vendors adopt the specification and new requirements appear, the community gr
 - [zeroheight](https://www.zeroheight.com)
 - ex-contributors to [Theo](https://github.com/salesforce-ux/theo) (by Salesforce)
 
+## Versions
+
+Here's an overview of all of the published versions of this draft specification:
+
+| name            | url                                      | date       |
```
drwpow, line 97:
```suggestion
| Name            | URL                                      | Date       |
```

README.md
```diff
@@ -90,6 +90,18 @@ As vendors adopt the specification and new requirements appear, the community gr
...
+| --------------- | ---------------------------------------- | ---------- |
+| editors-draft/1 | first-editors-draft.tr.designtokens.org  | 2021-09-23 |
```
drwpow, line 99:
```suggestion
| editors-draft/1 | [first-editors-draft.tr.designtokens.org](first-editors-draft.tr.designtokens.org)  | 2021-09-23 |
```

README.md
```diff
@@ -90,6 +90,18 @@ As vendors adopt the specification and new requirements appear, the community gr
...
+| editors-draft/2 | second-editors-draft.tr.designtokens.org | 2022-06-14 |
+| living-draft    | tr.designtokens.org                      | 2025-04-18 |
+
+> Note: tools can use the date as a version number to signify compliance. For example: `20250418`.
```
drwpow, line 103: I’d prefer `2025-04-18` instead because it’s more readable, and also screenreaders can announce that better than an integer. Also, having a string has the bonus feature of allowing more than one version per day, e.g. `2024-04-18-pre1` (which, you never know—always good to have the option)

drwpow (approved):
I think you’d want to get more reviews but this is a great addition IMO

c1rrus (approved):
Looks good. Thank you for adding this @jorenbroekema !

PR #286, opened 2025-07-20
feat: Add technical reports page

drwpow:
## Changes

Adds a new `/technical-reports/` URL that lists out all the published drafts. We don’t have a 3rd Editor’s Draft yet, but this could be included there when we have it.

CleanShot 2025-07-20 at 16 27 08@2x

CleanShot 2025-07-20 at 16 27 19@2x

## How to Review

- Manually tested
- See screenshots


www/_includes/assets/css/_global.scss
```diff
@@ -220,3 +220,33 @@ main {
...
   overflow-x: auto; // 2.
 }
+
+main table {
+  border-collapse: collapse;
+}
+
+main th {
+  border-bottom: 1px solid
+    color-mix(in oklch, #{$color-foreground}, transparent 75%);
```
drwpow, line 230: There wasn’t really a border token. Is this stupid?

ilikescience: if we can't do stupid things on our own website, what's the point of having a website?

ilikescience (approved):
looks good to me! i wonder if 'technical reports' is obviously the place to go to see the spec, but tbh it's the correct nomenclature and having something is better than having nothing and bikesheding this :)

PR #291, opened 2025-08-04
1. Improve Intro section

drwpow:
## Changes

Improves introduction section

## How to Review

- See comments

### DIFF
--- a/technical-reports/resolver/introduction.md
+++ b/technical-reports/resolver/introduction.md
@@ -1,10 +1,3 @@
 # Introduction
 
-Design tokens are key-value pairs that represent design decisions, allowing for consistent use across different platforms and tools. In complex design systems, there is a need to manage multiple dimensions such as themes, modes (e.g., light and dark), brands, and variants. This introduces challenges in organizing tokens, resolving references, and applying overrides.
-
-The Resolver Specification addresses these challenges by introducing a structured mechanism to:
-
-- Organize tokens into **sets** and **modifiers**.
-- Define how tokens are **resolved** when multiple sets and modifiers are involved.
-- Support **scalability** by handling arbitrary dimensions without complicating the core token definitions.
-- Maintain **modularity** and **reusability** by allowing tokens to be combined and overridden in a controlled manner.
+This module describes the means by which design tokens may express multiple values for the purposes of “theming” (light mode, dark mode, high contrast mode, etc.) or other scenarios.


technical-reports/resolver/introduction.md
```diff
@@ -1,10 +1,3 @@
 # Introduction
 
-Design tokens are key-value pairs that represent design decisions, allowing for consistent use across different platforms and tools. In complex design systems, there is a need to manage multiple dimensions such as themes, modes (e.g., light and dark), brands, and variants. This introduces challenges in organizing tokens, resolving references, and applying overrides.
```
drwpow, line 3: > Design tokens are key-value pairs that represent design decisions

Design tokens don’t need to be redefined, they have been defined in the top-level.

> allowing for consistent use across different platforms and tools.

This is an opinion statement, and the terms “platforms” and “tools” are not defined. Rather than define these, we can simplify the language.

> In complex design systems, there is a need to manage multiple dimensions such as themes, modes (e.g., light and dark), brands, and variants.

“There is a need” is also an opinion statement that does not describe why this exists. “Complex design systems” is also not described. What is a “design system?” What, quantitatively, defines it as “complex?”

Further, “themes,” “modes,” “brands,“ and “variants,” are all undefined, leading to further confusion. This section assumes prior knowledge without defining any of it, so it was removed.

technical-reports/resolver/introduction.md
```diff
@@ -1,10 +1,3 @@
...
-
-The Resolver Specification addresses these challenges by introducing a structured mechanism to:
```
drwpow, line 5: “The Resolver Specification” is incorrect, it should be “The Resolver module” (or something similar).

Further, we don’t need to summarize how these “challenges” are addressed because we never defined or described how these are problems to begin with, so the solutions are equally nebulous.

technical-reports/resolver/introduction.md
```diff
@@ -1,10 +1,3 @@
...
-
-- Organize tokens into **sets** and **modifiers**.
-- Define how tokens are **resolved** when multiple sets and modifiers are involved.
-- Support **scalability** by handling arbitrary dimensions without complicating the core token definitions.
-- Maintain **modularity** and **reusability** by allowing tokens to be combined and overridden in a controlled manner.
+This module describes the means by which design tokens may express multiple values for the purposes of “theming” (light mode, dark mode, high contrast mode, etc.) or other scenarios.
```
drwpow, line 3: Proposal: removing all opinions, undefined terms, and summary statements, this describes in plain language what this module does: define how multiple values for a single design token cna be expressed.

The only term introduced here is “theming,” but it is immediately followed by examples. If we want, we can define this further.

“Other scenarios” is a vague, undefined clause we may also define if we want; this could be improved.

technical-reports/resolver/introduction.md
mikekamminga: Agree on the opinion statement, we can probably rewrite these as use cases rather than specify for who the use case is?

drwpow: Yeah! To be clear I would _love_ for this information to exist! Just split out from the technical _specification_. An example is: if you look at any ARIA ReSpec specification, you’ll find few examples, no references to libraries, and the entire language is centered on painfully-boring details rather than possibities. But there’s a whole [Authoring Practices Guide](https://www.w3.org/WAI/ARIA/apg/) that goes in depth on examples. Anyways hopefully that’s clear! Only advocating for _where_ this info
should live, not actually changing any examples. We’re just only focused on the spec in this PR

drwpow:
Merging into the main resolver spec, so we can get better reviews internally next week! DTCG members can [see the internal discussion](https://design-systems.slack.com/archives/GPVT0EE9J/p1755011339740599)

PR #292, opened 2025-08-04
2. Terminology and 3. Syntax

drwpow:
## Changes

Improves Section 2. Terminology and 3. Syntax

## How to Review

- See comments


technical-reports/resolver/rationale.md
```diff
@@ -1,48 +0,0 @@
-## Rationale
-
-### Context and Problem Statement
-
-In design systems, especially those at scale, managing multiple themes, modes, brands, and other modifiers can become complex. Designers and developers often struggle with:
```
drwpow, line 5: This is a continuation of other points, but we are using terms without defining them, including `themes`, `modes`, `brands`, and `modifiers`. These terms, until defined, hold no meaning, therefore the following statements don’t hold weight.

> Designers and developers often struggle with:

This is an opinion statement not backed by data. It goes without saying it _motivated_ the specification, however, there are 2 reasons I’d like to remove it:

1. **It’s irrelevant to the introduction**. Even assuming this is true, this does not help readers understand the following content any better, and
2. **It‘s irrelevant to the specification.** These claims are not normative, and are irrelevant to any syntax or concepts following.

technical-reports/resolver/rationale.md
```diff
@@ -1,48 +0,0 @@
...
-
-Existing approaches, such as embedding modifiers directly into token names or using global theme files, have limitations. They often lead to:
-
-- **Global Namespace Pollution**: All tokens exist in a single global namespace, increasing the likelihood of naming conflicts.
-- **High Cognitive Load**: Designers and developers must remember complex naming conventions and how different tokens relate across dimensions.
-- **Performance Issues**: Resolving tokens in systems with many dimensions can be slow due to the combinatorial explosion of permutations.
-
-### Motivation
-
-The Resolver Specification aims to:
```
drwpow, line 30: A specification shouldn’t “aim to” accomplish anything. “Do or do not, there is no try.” etc. A specification should describe what it means to describe, and nothing more.

technical-reports/resolver/rationale.md
```diff
@@ -1,48 +0,0 @@
...
-- **Enhance Modularity**: Tokens can be organized into sets and modifiers, allowing teams to work on isolated components or themes without affecting others.
-- **Improve Performance**: By reducing the scope of tokens in resolution, the process becomes faster and more efficient.
-
-  <aside class="issue">
-
-  The term "process" needs clarification - what specific process is being referred to? Is this the "resolution process of the resolver"? More explanation is needed to understand exactly what becomes faster and more efficient when token scope is reduced during resolution.
-
-  </aside>
-
-- **Facilitate Tool Integration**: Providing a standard way to define resolution logic makes it easier for tools to integrate and automate token management.
```
drwpow, line 42: > Providing a standard way to define resolution logic

This is the entire purpose of specifications—to describe standard patterns of doing something.

technical-reports/resolver/rationale.md
drwpow: All the following bullet points are all unvalidated expectations of using this specification. However, I’d vote to remove it because they are unverifiable. We can’t measure any improvements without first outlining, at length, the original statuses of all cases. But also, doing so is beyond the purpose of any specification, which again, only describes syntax and format.

technical-reports/resolver/rationale.md
```diff
@@ -1,48 +0,0 @@
...
-
-In design systems, especially those at scale, managing multiple themes, modes, brands, and other modifiers can become complex. Designers and developers often struggle with:
-
-- **Naming Conflicts**: Using complex naming conventions to differentiate tokens across different themes or modes, leading to brittle and hard-to-maintain token names.
-- **Scalability Issues**: Difficulty in scaling token management when new dimensions are added, such as additional themes or brands.
-- **Tooling Limitations**: Existing tools may not support multi-dimensional token resolution, forcing teams to create workarounds that complicate their workflows.
-- **Separation of Concerns**: Embedding resolution logic within token files mixes data with behavior, violating separation of
-  concerns.
-
-### Parameters
```
drwpow, line 13: Removed because these are all **design goals,** not the specification itself. These may all be _motivators_ behind writing a specification! But writing down motivations alone is irrelevant to a specification.

technical-reports/resolver/terminology.md
```diff
@@ -1,103 +1,37 @@
-# Resolver Terminology
+# Terminology
 
-<dl>
-<dt>Resolver</dt>
-<dd>
+## <dfn>Resolver</dfn>
```
drwpow, line 3: Swapped from a definition list, which is NOT bookmarkable, to a simple heading hierarchy which is.

In ReSpec, adding `<dfn>…</dfn>` around anything allows it to be accessed with `[=someterm=]` throughout the document which is vital for reuse. We’ll make heavy use of this.

technical-reports/resolver/terminology.md
```diff
@@ -1,103 +1,37 @@
...
 
-A mechanism that combines token sets and applies modifiers to produce a final, resolved set of design tokens.
+The mechanism by which multiple possible values of design tokens are reduced to a single value, i.e. this module.
 
-</dd>
-<dt>Token Set</dt>
-<dd>
+## <dfn>Set</dfn>
 
-A collection of design tokens grouped together, often representing foundational tokens, component-specific tokens, or theme-specific tokens.
```
drwpow, line 13: > A collection of design tokens grouped together, often representing foundational tokens, component-specific tokens, or theme-specific tokens.

This is also an imprecise definition. “A collection of design tokens” is defined as a **group** already according to the spec.

Also, “foundational tokens,” “component-specific tokens,” etc. are **not definitions.** They are **examples!** Not an actual definition.

Improved all of these to try and form some language that communicates that these are _alternate_ design token values, intended to be merged in a specific order.

technical-reports/resolver/terminology.md
```diff
@@ -1,103 +1,37 @@
...
+Alternate design token values that are defined in [&lt;token-defs&gt;](#token-defs), intended to be merged in a specific order.
 
-</dd>
-<dt>Dimension</dt>
```
drwpow, line 16: “Dimension” was removed, because it is not only undefined, it also doesn’t appear in syntax in any way.

Open to keeping this around as a **non-normative definition,** but I think we should revisit how vital this concept is to a **specification** when it has no bearing on syntax or anything.

technical-reports/resolver/terminology.md
```diff
@@ -1,103 +1,37 @@
...
-<dd>
+## <dfn>Modifier</dfn>
 
-Dimensions are categories used to organize your design tokens. See them as contexts in which token values might change:
+Alternate design token value accessed with an [=input=], that takes priority over a [=set=]. Modifiers MAY be [=orthogonal=].
```
drwpow, line 13: Could probably use some help here 😅. I wanted to build off the idea of sets being **alternate versions of a design token** and also communicate these all have unique identifiers in a key–value relationship. They also take priority.

None of these may be _definitions,_ though, so thoughts welcome 🙏

technical-reports/resolver/terminology.md
```diff
@@ -1,103 +1,37 @@
...
-
-<dt>Modifier</dt>
-
-<dd>
-
-An entity that modifies or overrides tokens in the base sets. Modifiers can represent dimensions like themes, modes, brands, or any other contextual variations.
-
-<dl>
-
-<dt>Enumerated Modifier</dt>
```
drwpow, line 49: All the _types_ of modifiers here aren’t “terminology” per se, I think they should either be expressed as **Syntax** or outright removed.

technical-reports/resolver/syntax.md
```diff
@@ -0,0 +1,37 @@
+# Syntax
+
+## &lt;token-defs&gt;
```
drwpow, line 3: The whole concept of `<token-defs>` is something I am borrowing from other specs like [CSS Color Module 4](https://www.w3.org/TR/css-color-4/#normative) where a specific reoccurring syntax has a name.

The whole notion of an array of URIs + inline tokens appears frequently throughout the syntax, and warrants a definition that may be linked back to.

**I do not care about `<token-defs>` as a name!** As long as we give _whatever this is_ a name I think that’s sufficient.

technical-reports/resolver/syntax.md
```diff
@@ -0,0 +1,37 @@
+# Syntax
```
drwpow, line 1: This “Syntax” module will be fleshed out in future revisions.

At a higher level, the first cut of the Resolver spec outlines **too many examples** without breaking them down into their atomic parts. Instead, I’d like to:

1. **Define the atomic parts.** Describe the most granular aspects of the syntax, giving them all names
2. **Define how the atomic parts fit together.** Define the boundaries about how the syntax works as a cohesive whole.

I’m skipping ahead here, but the current process of “list out several examples” is **not a specification; it is a software library.** I would like to slow down and get “more in the weeds” with each part here.

All of this is above and beyond this PR, as this only adds this single entry under “Syntax.” But I’m describing “I would like more things to go here, and I know precisely what they are,” and have only put one thing here, now, for the sake of easier reviewing.

drwpow:
Resolving to move this forward and reviewable by next week! DTCG members can [see the related internal discussion](https://design-systems.slack.com/archives/GPVT0EE9J/p1755011339740599)

PR #293, opened 2025-08-16
chore(web): Upgrade eleventy to 3.x

drwpow:
## Changes

Updates eleventy to 3.x to work better with our ReSpec setup. This improves hot module reloading and updates all plugins to work better.

There were some insignificant changes as called out with comments

## How to Review

- See Netlify preview

PR #294, opened 2025-08-16
chore(www): Fix dev server refresh when editing respec

drwpow:
## Changes

Followup from #293, where auto refresh can now be fixed in eleventy. Editing the draft spec now reflects changes in the local server without a restart.

There were a few problems fixed:

1. Eleventy’s passthrough behavior, for some reason, disables auto-refreshing (other static site builders allow for asset refreshing just fine).
2. Eleventy also has special handling for HTML, so we have to basically treat `.html` as an “input” so Eleventy auto-reloads and auto-builds when it changes. But we have to disable processing and other automatic things that would cause issues for us.
3. There was also some magic behavior we had to opt out of, such as Eleventy respecting `.gitignore` files

There was an optional improvement made where now **only the spec you’re editing rebuilds in `dev` mode.** As opposed to any change rebuilding all specs.

## How to Review

- ✅ Locally, editing a spec now reflects changes in local server (takes a few seconds for ReSpec to build, but that’s expected)
- ✅ The Netlify preview was expected, and the changes to the Eleventy config didn’t cause any regressions on the site

### DIFF
--- a/technical-reports/package.json
+++ b/technical-reports/package.json
@@ -10,10 +10,14 @@
   },
   "scripts": {
     "build": "pnpm run build:index && pnpm run build:format && pnpm run build:color",
-    "build:index": "mkdir -p ../www/TR/drafts && respec index.html ../www/TR/drafts/index.html --localhost --disable-sandbox",
-    "build:format": "mkdir -p ../www/TR/drafts && mkdir -p ../www/TR/drafts/format && respec format/index.html ../www/TR/drafts/format/index.html --localhost --disable-sandbox",
-    "build:color": "mkdir -p ../www/TR/drafts && mkdir -p ../www/TR/drafts/color && respec color/index.html ../www/TR/drafts/color/index.html --localhost --disable-sandbox",
-    "dev": "chokidar \"**/*.{html,md}\" -c \"pnpm run build\" -d 2000",
+    "build:index": "pnpm run mkdirs && respec index.html ../www/TR/drafts/index.html --port 3001 --localhost --disable-sandbox",
+    "build:format": "pnpm run mkdirs && respec format/index.html ../www/TR/drafts/format/index.html --port 3002 --localhost --disable-sandbox",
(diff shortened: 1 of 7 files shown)


technical-reports/package.json
```diff
@@ -10,10 +10,14 @@
...
-    "build:color": "mkdir -p ../www/TR/drafts && mkdir -p ../www/TR/drafts/color && respec color/index.html ../www/TR/drafts/color/index.html --localhost --disable-sandbox",
-    "dev": "chokidar \"**/*.{html,md}\" -c \"pnpm run build\" -d 2000",
+    "build:index": "pnpm run mkdirs && respec index.html ../www/TR/drafts/index.html --port 3001 --localhost --disable-sandbox",
+    "build:format": "pnpm run mkdirs && respec format/index.html ../www/TR/drafts/format/index.html --port 3002 --localhost --disable-sandbox",
+    "build:color": "pnpm run mkdirs && respec color/index.html ../www/TR/drafts/color/index.html --port 3003 --localhost --disable-sandbox",
+    "mkdirs": "mkdir -p ../www/TR/drafts && mkdir -p ../www/TR/drafts/format && mkdir -p ../www/TR/drafts/color",
+    "dev": "pnpm --parallel --filter @dtcg/tr run \"/^dev:.*/\"",
+    "dev:index": "pnpm run mkdirs && chokidar \"index.html\" -c \"pnpm run build:index\" -d 5000",
+    "dev:format": "pnpm run mkdirs && chokidar \"format/**/*\" -c \"pnpm run build:format\" -d 5000",
+    "dev:color": "pnpm run mkdirs && chokidar \"color/**/*\" -c \"pnpm run build:color\" -d 5000",
```
drwpow, line 20: Speedup: now only the spec you’re actively editing refreshes, saving time

drwpow:
Also going to merge this without a review, as it’s also another chore that has no impact on any specifications, presentations, content, or decisions. 100% maintenance.

PR #297, opened 2025-08-17
Remaining edits to the resolver specification

drwpow:
## Changes

Last major PR to get the resolver specification reviewable internally. At a high level this fixes:

- Adds incomplete or missing definitions
- Fixes examples that are invalid DTCG format or pseudocode
- Renames `meta.alias` → `meta.namespace` just for review. The term “alias” was double-loaded with two meanings so we had to use a different one. We can go with a different term ultimately!
- Deduplicates and reorganizes sections to avoid repeating information

There are some more nuanced changes that will be outlined in comments.

## How to Review

- See comments!


technical-reports/resolver/additional-context-and-information.md
```diff
@@ -1,78 +0,0 @@
...
-}
-```
-
-Resulting in tokens accessible via spacing.sm and spacing.lg.
-
-</aside>
-
-<aside class="example">
-
-### Real-World Use Case: GitHub Primer
```
drwpow, line 54: For now, I removed this block since this is merely a reference to a design system, without clear relevance to the specification. It didn’t really come up in other examples.

technical-reports/resolver/example3.md
```diff
@@ -1,283 +0,0 @@
-# Example 3: Responsive Design
```
drwpow, line 1: See comment on [resolution-logic.md](https://github.com/design-tokens/community-group/pull/297/files#diff-0e6cd111dbd258bfa385d34d4186657dcd2df27af6b19dc41baa37054c5014d4)

technical-reports/resolver/resolution-logic.md
```diff
@@ -1,94 +1,416 @@
...
-  - Apply any aliasing or namespacing specified in meta.
-  - Load the token sets associated with the selected modifier value.
-  - Merge these tokens with the base tokens, applying overrides as necessary.
+## Resolution
 
-#### Alias Resolution
+After all aliases resolve correctly in the final set, the end result is one tokens object, that behaves as if it was a single JSON file to begin with.
 
-<aside class="note">
+<aside class="example" title="Theme resolution">
```
drwpow, line 198: Examples 1, 2, and 3 were removed in favor of this single example. This one example showcases all resolution logic steps, and namespacing, without overcomplicating it.

Example 1, 2, and 3 do not introduce any new concepts—all use the same resolution logic steps and namespacing. They do not describe non-orthogonal resolution, invalid specs, error handling, or any other different scenario that illustrates a new concept.

If we could modify any of those examples to illustrate a new concept we could re-add them. But the motivating factor should be illustrating unique syntax, and not merely duplicating other examples with different values.

technical-reports/resolver/index.html
```diff
@@ -48,12 +48,9 @@
...
       <p>
-        This document describes the technical specification for a resolver
-        mechanism used to manage and resolve design tokens in complex scenarios
-        involving themes, modes, brands, and other modifiers. It extends the
-        Design Tokens Format Specification by introducing a structured way to
-        combine and override token sets to produce a final, resolved set of
-        tokens for consumption by design tools and platforms.
+        This specification extends the [format](../format/) and describes a
+        method to work with alternate values for [design tokens](../), such as
+        “light mode” and “dark mode” color themes for supporting devices.
```
drwpow, line 53: Was reading the [W3C manual of style](https://www.w3.org/guide/manual-of-style/#Status) and ran across this:

> Tim Berners-Lee expressed the goal of the custom paragraph this way, "Don't be afraid of being honest about the relevant techno-political situation." In the custom paragraph, make the case for why someone should read this draft.

In light of that, I wanted to take a crack at providing a clearer, more succinct way of describing this. This is an opinionated revision! And I would be open to changing this to better follow the styleguide

technical-reports/resolver/introduction.md
```diff
@@ -1,3 +1,105 @@
...
+<aside class="example" title="Naïve token increase">
+
+Starting with 100 [color tokens](../format/#color):
+
+- Adding a 2nd “dark mode” layer of alternate values results in 200 values, or 100 × 2.
+- Adding a 2nd (“dark mode”), 3rd (“light mode - high contrast”), and 4th (“dark mode - high contrast”) layer results in 400 values, or 100 × 4.
+
+</aside>
+
+## Resolver approach
```
drwpow, line 28: Image

This is also the only real opinionated addition I had here. Previous revisions had examples that required the reader to have read many blog posts and/or lectures on the topic. I was not able to find an explanation of the core problem that required no external knowledge.

I may have gone overboard on the math here, but I wanted to illustrate the approach and value here by just comparing numbers. In this example, we compare 2 approaches, one where you end up with 2,000 tokens not using the resolver method, and the other where you end up with 700. I know math isn’t always “approachable” but I was hoping this would be an illustration that communicated the concept without having to lean on abstract concepts.

technical-reports/resolver/orthogonality.md
```diff
@@ -1,26 +0,0 @@
-# Orthogonality
```
drwpow, line 1: Reorganized to live inside `syntax.md`, under “Modifiers”

technical-reports/resolver/resolution-example.md
```diff
@@ -1,199 +0,0 @@
-# Resolver Resolution Example
```
drwpow, line 1: This example was kept and just moved as the primary example under `resolution-logic.md`

PR #301, opened 2025-08-28
Resolver spec updates

drwpow:
## Changes

This makes 2 significant changes to **syntax** and **removing `namespacing`**.

### Syntax

The syntax was changed from `sets` and `modifiers` being declared separately, to being merged in the same array. This solves the following problems:

- Modifier types (enumerated and include) are no longer needed, because the “include” modifier was really just a set in disguise (a set declared after a modifier now serves the same function)
- If someone wanted to include a token file that override a modifier, that was impossible before.
- Boilerplate is reduced significantly
- No more contextual `name` and `values` being used in both sets and modifiers. Now sets and modifiers have distinct syntax that are easy to scan.

The changes seem significant, but are backwards-compatible. In many ways this is just a cleanup.

### Namespace removal 

Namespacing tokens doesn’t have a clear usecase, rather, introduces complexity in the specification.

- The existing DTCG format relies on a **global token schema** that is consistent across the design system. This disrupts that by making the global token schema unknowable.
- The responsibility of naming is now split between resolver and token files, with neither  having complete control. Aliases become unusable if a modifier, at any stage in the process, is able to break the link. In other words, it’s a circular dependency.
- By removing namespacing, the naming responsibility and schema is now 100% back on tokens, with the resolver mostly resolving conflicts. 

## How to Review

Discussed in this week’s DTCG meeting; see [notes](https://docs.google.com/document/d/1sfiVmnIxgeiA20JJ4IOkm8da_CM9hTnsL2gpzO8vEOM/edit?tab=t.0#heading).

### DIFF
--- a/technical-reports/resolver/resolution-logic.md
+++ b/technical-reports/resolver/resolution-logic.md
@@ -3,11 +3,8 @@
 Tools MUST handle the resolution stages in this order to produce the correct output.
 
 1. [Input validation](#input-validation)
-2. [Set flattening](#set-flattening)
-3. [Modifier application](#modifier-application)
-4. [Namespacing](#namespacing)
-5. [Alias resolution](#alias-resolution):
-6. [Resolution](#resolution-0)
+2. [Token flattening](#token-flattening)
+3. [Resolution](#resolution-0)
 
 ## Input validation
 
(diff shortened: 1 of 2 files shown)


technical-reports/resolver/resolution-logic.md
```diff
@@ -21,48 +18,35 @@ If a resolver does NOT declare any modifiers, skip this step and proceed to [Set
...
+1. Every token reference in `sources` MUST be resolved in array order.
+1. For every token reference, merge with the previous set.
+1. [Aliases](../format/#aliases-references) MUST NOT be resolved yet. That must happen at the end. Aliases MAY refer to values that will be supplied in upcoming steps.
 
 <aside class="example" title="Conflict resolution">
 
 ```json
 {
-  "sets": [
+  "tokens": [
```
drwpow, line 37: `"tokens"` is a placeholder, we’re not really crazy about and is subject to change. We just need “something” here and anything will suffice.

technical-reports/resolver/resolution-logic.md
```diff
@@ -21,48 +18,35 @@ If a resolver does NOT declare any modifiers, skip this step and proceed to [Set
...
     {
+      "type": "set",
       "name": "foundation",
-      "values": [
+      "sources": [
```
drwpow, line 41: `"sources"` is much the same—it was suggested in #289 as a better alternative to “value” which has meaning in the DTCG format.

technical-reports/resolver/resolution-logic.md
```diff
@@ -137,23 +122,48 @@ Here, 2 `color.text.default` tokens were supplied, one after the other. Since th
...
+
+</aside>
+
+<aside class="example" title="Simple syntax comparison">
+
+All of the following MUST be flattened in the same order, and are equivalent to one another:
+
+```json
+{
+  "tokens": ["base/foundation.json", "base/colors.json"]
```
drwpow, line 135: This is the motivation for the change—in a nutshell. This is a complete resolver! 🎉. It can accomplish all the same things as before, just in a user-friendly package.

PR #303, opened 2025-09-02
fix color property on shadow, closes #270

### LINKED ISSUE #270: Color property for shadow token is inconsistent with the rest.
Color property is defined as a token while the rest are not (see https://tr.designtokens.org/format/#example-shadow-token-example)

```json
{
  "shadow-token": {
    "$type": "shadow",
    "$value": {
      "color": {
        "$type": "color",
        "$value": {
          "colorSpace": "srgb",
          "components": [0, 0, 0],
          "alpha": 0.5,
          "hex": "#000000"
        }
      },
      "offsetX": { "value": 0.5, "unit": "rem" },
      "offsetY": { "value": 0.5, "unit": "rem" },
      "blur": { "value": 1.5, "unit": "rem" },
      "spread": { "value": 0, "unit": "rem" }
    }
  }
}
```

Shouldn't it be

```json
{
  "shadow-token": {
    "$type": "shadow",
    "$value": {
      "color": {
        "colorSpace": "srgb",
        "components": [0, 0, 0],
        "alpha": 0.5,
        "hex": "#000000"
      },
      "offsetX": { "value": 0.5, "unit": "rem" },
      "offsetY": { "value": 0.5, "unit": "rem" },
      "blur": { "value": 1.5, "unit": "rem" },
      "spread": { "value": 0, "unit": "rem" }
    }
  }
}
```

ilikescience:
## Changes

Updates examples in shadow composite type definition to be valid. Previously, it used an entire _key/value pair_ as the value of the 'color' property, whereas according to the spec it should be a _color value_.

Closes #270 — I will submit a followup to fix the other examples where we're using a color token instead of a value in a composite token (border, gradient, etc).

## How to Review

Normal review, check the spec.

### DIFF
--- a/technical-reports/format/composite-types.md
+++ b/technical-reports/format/composite-types.md
@@ -323,84 +323,129 @@ Represents a shadow style. The `$type` property MUST be set to the string `shado
     "$type": "shadow",
     "$value": {
       "color": {
-        "$type": "color",
-        "$value": {
-          "colorSpace": "srgb",
-          "components": [0, 0, 0],
-          "alpha": 0.5,
-        }
-      },
-      "offsetX": { "value": 0.5, "unit": "rem" },
-      "offsetY": { "value": 0.5, "unit": "rem" },
-      "blur": { "value": 1.5, "unit": "rem" },
-      "spread": { "value": 0, "unit": "rem" }
-    }
-  },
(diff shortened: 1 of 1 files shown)


drwpow (approved):
Thank you!

PR #313, opened 2025-10-05
chore: skip www build on dev command

drwpow:
## Changes

Chore. Speeds up `pnpm run dev` by skipping a full `astro build` (this was accidentally left over from the 11ty setup).

Also updates husky to use `pnpm exec` instead of npx, and updates those deps. This removes some local errors when developing, and speeds up the githook.

## How to Review

- These are only local changes, but can be tested locally if desired. These don’t affect the prod site.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,1 +1,1 @@
-npx lint-staged
+pnpm exec lint-staged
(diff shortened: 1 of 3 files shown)


drwpow:
I’m going to merge this as this is unopinionated dev cleanup, will get review on the other one

PR #314, opened 2025-10-06
feat: add md plugins

drwpow:
## Changes

Adds very minor, lazy style cleanup, and changes the primary color from gold → blue. I think we could probably improve this a little, it just simplifies the design a little in preparation for a real one.

Primarily, though, this adds:

- Heading anchor links to each h# tag
- Table of contents for markdown-generated pages
- Minor improvements to `:focus-visible` outlines for a11y

Beyond that, I don’t feel strongly about any design choices made. More just wanted to prepare it for followup iterations.

**Screenshots**

CleanShot 2025-10-05 at 20 41 32@2x

CleanShot 2025-10-05 at 20 41 43@2x

CleanShot 2025-10-05 at 20 42 08@2x

https://github.com/user-attachments/assets/569b9e00-8039-4fac-a471-0c354f2886d8

## How to Review

- See screenshots
- See Netlify deploy


www/md-plugins/rehype-auto-toc.js
```diff
@@ -0,0 +1,83 @@
+export default function transform() {
```
drwpow, line 1: This is just a fork of a fork of a custom markdown plugin I have lying around. The Starlight one is too deeply-tied to Starlight internals. I don’t know of another common one that’s easy-to-use. But when you look at how they’re made, they’re actually really really simple, as you can see here it’s < 100 lines, and all it’s doing is just crawling the markdown AST once and making a nested `<ol>` out of it.

www/src/layouts/Markdown.astro
```diff
@@ -0,0 +1,87 @@
...
+      }
+
+      h6 {
+        @include text-heading6;
+      }
+    }
+  </style>
+  <script src="@/scripts/toc"></script>
+
+  <div class="md-content">
```
drwpow, line 84: It’s probably a little hard to see the diff here, but in a nutshell:

1. `Layout.astro` → `Base.astro`
2. The markdown things were moved into `Markdown.astro`
3. `Markdown.astro` extends `Base.astro` with more things

The reason is some pages like `blog` don’t involve markdown content at all, and some of the styling was a little weird between a tightly-managed content section, vs a markdown content page. The goal is that custom pages use a simpler layout, and the markdown pages use more complete styling

www/src/components/Footer.astro
```diff
@@ -0,0 +1,137 @@
...
+                </a>
+              </li>
+            ))
+          }
+        </ul>
+      </nav>
+    </div>
+  </div>
+
+  <style lang="scss">
```
drwpow, line 72: @ChucKN0risK you can tell me to revert this, it’s fine. I just struggled already with a) making sure my classnames between components don’t have any duplicates/conflicts, and b) wanting to be lazy about naming without it bleeding into the global scope. The automatic scoping in each `.astro` component is just really good

www/src/components/SocialIcon.astro
```diff
@@ -0,0 +1,122 @@
...
+  url: string;
+}
+
+const { url, ...props } = Astro.props;
+
+type Icon = 'github' | 'opencollective' | 'linkedin' | 'x' | 'w3c' | 'bluesky';
+
+let icon: Icon | undefined;
+
+if (url.includes('bluesky.app') || url.includes('bsky.app')) {
```
drwpow, line 12: URL sniffing is very dumb; it was only out of laziness to reuse the existing nav links without adding another property

www/src/assets/css/01-utils/tokens/_colors.scss
```diff
@@ -1,4 +1,269 @@
 :root {
+  // Tailwind 4.0 colors
```
drwpow, line 2: I know it’s lame to use the Tailwind colors, but they have a really even balance and I was too lazy to find a better ramp that works well in light/dark mode (I’m not really a fan of Radix, and a lot of other systems IMO don’t really provide a nice spread)

www/src/components/Footer.astro
```diff
@@ -0,0 +1,172 @@
...
+          menu.about.map((link) => (
+            <li>
+              <a
+                aria-current={url.pathname === link.path ? 'page' : false}
+                data-astro-prefetch
+                class="link"
+                href={link.path}
+                aria-label={
+                  isLinkAbsolute(link.path)
+                    ? `Link opens in a new window (${link.path})`
```
drwpow: cc @kaelig I have to make an adjustment here because `aria-label` _replaces_ all meaningful text within, not merely appending. So as-written, screen readers would _only_ see “Link opens in a new window ([link path])” which isn’t a good accessible name (also they don’t need the address read out loud which could be noisy). `aria-description` is better for supplementary info that doesn’t replace the accessible name.

But for a simpler approach, I’m just giving the `<Icon />` inside a description that it’s an external link. That way the screen reader experience is 1:1 with the element structure

www/src/pages/blog.astro
```diff
@@ -24,29 +22,40 @@ const formattedPosts = sortedPosts.map((post) => {
...
-            <Text as='h3' variant='heading3'>
-              <a href={post.url} aria-label={`Read ${post.title}`}>
-                {post.title}
-              </a>
-            </Text>
-            <Text as='p' variant='small'>
+          <Stack as="article" gap={2} class="article">
+            <a
+              href={post.url}
+              aria-label={`Read ${post.title}`}
```
drwpow: Also removed this aria-label override. Adding additional words for screen readers when there’s no visual indicator isn’t necessary. Screen reader users should just have the links read the same as sighted users. This creates unnecessary repetition.

PR #345, opened 2025-10-28
feat: add social image

drwpow:
## Changes

Adds a social image to embeds from Tim’s design

social

## How to Review

- Are the meta tags correct?


julien-deramond:
> Are the meta tags correct?

Sorry to jump in. I’m subscribed to this repo’s notifications to follow DTCG updates, but I noticed something familiar from when we migrated Bootstrap’s docs to Astro.

FYI, there’s a fallback mechanism on Twitter/X’s side that relies on `og:*` tags: https://developer.x.com/en/docs/x-for-websites/cards/overview/markup

We ended up using the following config in Bootstrap, which might help: https://github.com/twbs/bootstrap/blob/main/site/src/components/head/Social.astro

PR #349, opened 2025-11-11
feat: add DTCG playground

drwpow:
## Changes

Adds [token playground](https://deploy-preview-349--designtokensorg.netlify.app/playground/) at `/playground` ([original source](https://github.com/terrazzoapp/dtcg-resolver-playground)). Adds 2 examples—GitHub Primer and Figma SDS, both MIT-licensed.

https://github.com/user-attachments/assets/a195963b-7479-412f-9e81-5c580007e27e

Submitting to this repo licenses all code under the existing W3C license.

## How to Review

- [**View Playground**](https://deploy-preview-349--designtokensorg.netlify.app/playground/)
- See comments
- This only adds a `/playground` URL for us to preview. Doesn’t add link anywhere

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,11 +26,11 @@
     "*.{md,yml,json,html}": "prettier --write"
   },
   "devDependencies": {
-    "cspell": "^9.3.2",
+    "cspell": "^9.4.0",
     "husky": "^9.1.7",
     "lint-staged": "^16.2.7",
-    "prettier": "^3.6.2",
+    "prettier": "^3.7.4",
     "prettier-plugin-astro": "^0.14.1",
-    "puppeteer": "^24.31.0"
+    "puppeteer": "^24.32.0"
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     devDependencies:
       cspell:
-        specifier: ^9.3.2
-        version: 9.3.2
+        specifier: ^9.4.0
+        version: 9.4.0
       husky:
         specifier: ^9.1.7
         version: 9.1.7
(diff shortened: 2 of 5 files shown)


.zed/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "formatter": "prettier"
```
drwpow: I’ve been using Zed more often; we have a `.vscode` folder so hopefully this isn’t too much noise?

www/src/assets/css/02-base/_global.scss
```diff
@@ -87,14 +87,14 @@ pre {
   overflow-x: auto;
 }
 
-main table {
+.page table {
```
drwpow: Change: don’t add global styling for all `<main>` elements

www/src/components/TokenPlayground/lib/examples/figma-sds.ts
```diff
@@ -0,0 +1,1565 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// JSON.stringify() stores these in memory as strings, but declaring them as
+// objects prevents simple typos and errors
+export default {
+  'sds.resolver.json': JSON.stringify({
```
drwpow: Note on these files: these are in `.ts` files because they’re way more portable. Ironically, if these were `.json` files, we’d have to make sure everything is imported and bundled correctly, which is more work (because we probably wouldn’t want 1 network request per `.json` file, we’d be bundling, but there’d be more indirection in the setup). Since we need all these at runtime, and the user is editing in-browser, just doing the work of putting them in a TypeScript object is more straightforward.

kaelig (approved):
Still getting a vertical scrollbar, but that can be fixed later!
Screenshot 2025-12-05 at 12 52 18 AM

PR #356, opened 2025-11-22
feat: Add JSON Schema for resolver documents

drwpow:
## Changes

CleanShot 2025-11-21 at 09 32 01@2x

Adds a public JSON Schema for the resolver specification for more user-friendly access. Also adds to examples in the specification, without changing any behavior.

Also adds a brief explanation of what it’s used for, also without changing any behavior or syntax.

## How to Review

See preview URL

### DIFF
--- a/www/public/schemas/2025.10/resolver.json
+++ b/www/public/schemas/2025.10/resolver.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "schemas/2025.10/resolver.json",
+  "$id": "https://www.designtokens.org/schemas/2025.10/resolver.json",
   "title": "DTCG Resolver Schema",
   "description": "JSON Schema for the Design Tokens Community Group (DTCG) Resolver specification",
   "type": "object",
--- a/www/public/schemas/2025.10/format.json
+++ b/www/public/schemas/2025.10/format.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "schemas/2025.10/format.json",
+  "$id": "https://www.designtokens.org/schemas/2025.10/format.json",
   "title": "DTCG Format Schema",
   "description": "JSON Schema for the Design Tokens Community Group (DTCG) Format specification.",
   "type": "object",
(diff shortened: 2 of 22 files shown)


www/public/schemas/2025.10/resolver.json
```diff
@@ -0,0 +1,155 @@
+{
+  "type": "object",
```
drwpow: Note: I tested this schema locally, and works as-expected in VS Code. Some other editors may have different levels of how much they support, for example Zed doesn’t respect `oneOf` but VS Code does.

I’m treating this as a “better than nothing” approach. If there are any bugs, we can fix them as they are reported.

dgtlntv:
Sure I would be happy to create a PR! :) With the tests I created and with my own usage it seemed to work well, also for arbitrarily nested groups/tokens. The way I have done it is in the group definition I allow any property name that is a valid group or token name and has as its value allowed group or token values like so (can be found in the [group.json file](https://github.com/dgtlntv/dtcg-json-schema/blob/main/schemas/format/group.json)):

```json
"patternProperties": {
    "^[^${}.][^{}.]*$": {
        "description": "Nested groups and tokens (see ./format.json#/$defs/tokenOrGroupName for pattern definition)",
        "$ref": "https://dgtln.tv/schemas/2025.10/format/groupOrToken.json"
    }
}
```

Is that what you meant or something else? But I might have just not tested it well enough yet. If you have an example of what has been challenging to validate with a JSON schema in the past I'd be happy to make sure that it works!

I also just saw that schemastore recommends using the `draft-07` JSON schema version. I have used the `draft-2020-12` version in my schema. I haven't yet checked wether my schema is also compliant with the `draft-07`. So depending on what version we want to target I might also need to update things in the schema and maybe some limitations arise from that.

drwpow:
Closing in favor of #360

ilikescience (approved):
lgtm! i'm no pro user of schemas, so i tested with `ajv` — there's some warnings in strict mode that we may want to handle, but i leave it to the author to decide if that's important or not.

drwpow:
That’s awesome! And interesting, hm. The non-bundled version seemed to be working fine for me, but I wonder if the bundling changed something? 🤔 

Also on that thread you linked, this comment caught my attention: a bundler that works around VS Code limitations: https://github.com/microsoft/vscode-json-languageservice/issues/224#issuecomment-2174499089. Dunno if it has drawbacks based on what you did though 🤷

PR #357, opened 2025-11-22
Fix version to be 2025.10

drwpow:
## Changes

Follow-up from #352 where a placeholder version slipped through final review, and the resolver module published with `YYYY-MM-DD` version instead of `YYYY.MM` version as originally-intended. Fortunately this has no current bearing on any tool behavior, but nonetheless we should correct it ASAP.

## How to Review

See diffs


drwpow:
Going to merge and release without approval, just because this will minimize mistakes the faster it’s fixed.

Issue #249, opened 2024-09-10
Is the root node of a token tree a standard group?


nclsndr:
I couldn't find this detail in the current spec: Do we consider the root node of a token tree a standard group?
Said differently, can I have `$description` and/or `$extensions` at the root level? 👇 

```json
{
  "$description": "Tokens from brand A",
  "color": {
    "$type": "color",
    "$value": "#ff0000"
  }
}
```

---

It seems giving the root node the group definition opens opportunity to store global metadata. But becomes harder to work with, anytime there're more than 1 tokens file to handle.

Is there a decision about this already?

drwpow:
This isn’t explicit, you’re right, but the root node is a group node. Any node that isn’t a token node (contains `$value`) is a group node. So any properties allowed on group are allowed at the root, and will cascade as such.

There have been some proposals to add root properties (such as `$name` to name the file), which would change this, but none are part of the spec currently.

Out of convention, though, most people don’t place anything in the root because there’s rarely any properties that make sense to apply to all tokens (`$description`, `$type`), but there’s nothing _explicitly_ preventing this. Both of your examples are valid.

Technically, too, the root node _could_ be a token if it contained `$value`. While that’s technically valid, it’s impractical, because no one has a practical use for a design system with only one token.

nclsndr:
Thx @drwpow for the clear answer 👍

nclsndr:
I was looking into the spec to slide in this bit of information. But I find myself not knowing where it would fit? Anyone with opinion?

drwpow:
> I was looking into the spec to slide in this bit of information. But I find myself not knowing where it would fit? 

I think this would be a good addition to the [Group](https://tr.designtokens.org/format/#groups) section! In comparison to the **Design Token** section which outlines that a token MUST have the `$value` attribute, it’s assumed that any other object that doesn’t have `$value` (and isn’t inside a `$value`) is a Group node. But you’re right this isn’t explicated.

Issue #247, opened 2024-08-20
How library and tooling authors do keep track of conformance and changes?


nclsndr:
Lately, I reviewed and benchmarked many of the OSS solutions for working with DTF tokens (Tokens Studio, Cobalt/Terrazzo, Design Token Validator from Anima,...). Side note: I focused on web tech, JS and TS.

What struck me was the level of repetition we all suffer from. 100% of any lib that parses DTF tokens needs to start by defining cumbersome types and primitives. Then parse and implement some sort of visitor pattern for both the tokens and their aliases. I can tell you we almost all have made mistakes interpreting some decisions or missing an update from the spec at some point 🙈.

Taking this step back, I feel like we would all benefit from a npm package that would host the types and the most implementation-agnostic definitions. I wrote the package [design-token-format-module](https://github.com/nclsndr/design-tokens-format-module) in late 2022, by that time it was not clear where it could be used and what type of APIs it could offer. 
With the last version, targeting the Live Draft spec, I tried to focus on the most non-debatable piece for anyone to pick from.

Questions for library/tool authors: 
- How do you envision to keep track of the spec changes and conformance in your codebase(s)?
- Do you feel a shared package of primitives would help?

Question for DTCG committee:
Would it make sense for the design-tokens organization on GitHub to host some repositories with the code implementing the spec for different languages / platforms? If so, I'd be happy to transfer the npm namespace to the community.

drwpow:
> What struck me was the level of repetition we all suffer from.

I’m just asserting my opinion here, but I think ideally this is how Open Source works? Having more implementations, and more options, only benefit users.

Take YAML, for instance. It has a central specification defined of how it _should_ work. But the group that wrote that specification aren’t responsible for porting it into every programming language; that’s done by hundreds (if not thousands) of individual developers. The end result is, yes, there’s a lot of “repetition.” But each individual implementation works extremely well for the users it’s serving.

I always view this as a feature, not a bug. Anyone being able to create their own library and open source it is always a great thing. And the community is always benefitted by _more_ open source projects, not fewer. OSS is a lot of work! And there’s a real resource cost to trying to centralize implementation and put the burden on the standards committee, rather than distributing the load across developers, with many people being able to do part of the lifting.

But that may just be my viewpoint—I just don’t see GitHub projects “competing” with one another. I all view it under the same umbrella of collaboration and mutual sharing. That is, as long as it’s open-licensed and there’s not money/profit involved 😉 

[...]

PR #360, opened 2025-12-01
Add format schema

dgtlntv:
As discussed in #356 I am creating a PR to contribute the format JSON schema [that I created](https://github.com/dgtlntv/dtcg-json-schema). 

Open questions:
- This schema allows $schema at the root of a design token file while the current technical reports does not allow for it. Should this be kept or removed?
- This schema assumes that the root of a design token file is a group as discussed in #249. Is that assumption correct?
- This schema expects 3 components for all color values, though the specification states "the number of components depends on the color space." Should it be made more accurate to the spec or kept like this for now?
- This schema is currently written for the `draft-07` JSON schema version. Is that the version that should be targeted?
- Should the tests I have written for the schema also be included in the PR or just the schema?
- Does this schema work with the arbitrarily-nested DTCG files issue mentioned in #356?

## How to Review
- Test this schema with correct and incorrect design token files and check wether it accurately validates the Design token files.

### DIFF
--- a/www/public/schemas/2025.10/format.json
+++ b/www/public/schemas/2025.10/format.json
@@ -1,6 +1,6 @@
 {
   "$schema": "http://json-schema.org/draft-07/schema#",
-  "$id": "schemas/2025.10/format/format.json",
+  "$id": "schemas/2025.10/format.json",
   "title": "DTCG Format Schema",
   "description": "JSON Schema for the Design Tokens Community Group (DTCG) Format specification.",
   "type": "object",
@@ -12,7 +12,7 @@
       "$comment": "$schema is not part of the official DTCG specification."
     },
     "$type": {
-      "$ref": "tokenType.json",
+      "$ref": "format/tokenType.json",
       "description": "The type for tokens in this group (inherited by nested tokens unless overridden)"
     },
     "$description": {
@@ -47,13 +47,13 @@
     },
     "$root": {
       "description": "Root token for this group. The $root token provides a base value for the group while allowing for variants or extensions.",
-      "$ref": "token.json"
+      "$ref": "format/token.json"
     }
   },
   "patternProperties": {
     "^[^${}.][^{}.]*$": {
       "description": "Nested groups and tokens (see #/definitions/tokenOrGroupName for pattern definition)",
-      "$ref": "groupOrToken.json"
+      "$ref": "format/groupOrToken.json"
     }
   },
   "additionalProperties": false,
--- a/www/public/schemas/2025.10/format/group.json
+++ b/www/public/schemas/2025.10/format/group.json
@@ -20,10 +20,10 @@
     "$extends": {
       "oneOf": [
         {
-          "$ref": "format.json#/definitions/curlyBraceReference"
+          "$ref": "../format.json#/definitions/curlyBraceReference"
         },
         {
-          "$ref": "format.json#/definitions/jsonPointerReference"
+          "$ref": "../format.json#/definitions/jsonPointerReference"
         }
       ],
       "description": "Reference to another group to inherit tokens and properties from"
@@ -46,7 +46,7 @@
   },
   "patternProperties": {
     "^[^${}.][^{}.]*$": {
-      "description": "Nested groups and tokens (see ./format.json#/definitions/tokenOrGroupName for pattern definition)",
+      "description": "Nested groups and tokens (see ../format.json#/definitions/tokenOrGroupName for pattern definition)",
       "$ref": "groupOrToken.json"
     }
   },
--- a/www/public/schemas/2025.10/format/token.json
+++ b/www/public/schemas/2025.10/format/token.json
@@ -20,7 +20,7 @@
       "$ref": "tokenType.json"
     },
     "$ref": {
-      "$ref": "format.json#/definitions/jsonPointerReference",
+      "$ref": "../format.json#/definitions/jsonPointerReference",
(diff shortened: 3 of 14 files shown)


www/public/schemas/2025.10/format/values/fontWeight.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://designtokens.org/schemas/2025.10/format/values/fontWeight.json",
+  "title": "Font Weight Value",
+  "description": "Value schema for fontWeight type tokens. Represents a font weight as per the OpenType wght tag specification. Lower numbers represent lighter weights, higher numbers represent thicker weights.",
```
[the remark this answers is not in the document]
drwpow: Yeah that makes sense! The spellcheck, as you found, is a bit of a double-edged sword—it has caught a lot of spelling mistakes, but at the cost of being annoying like this. There’s also some subjectivity, too—if we think a term is only used in one place we can add an “ignore” comment. But `wght` seems like a) we’d use it in multiple places, and b) it’s not a common misspelling someone would likely make.

So short answer: yes we could add it to the “terminology” list (the separation is arbitrary; it just makes each list a little easier to sort through with logical groups).

drwpow:
> * This schema allows $schema at the root of a design token file while the current technical reports does not allow for it. Should this be kept or removed?

I think we’ll likely revisit the language of the spec. But I think we’d need to keep it, wouldn’t we? Otherwise this couldn’t exist! I think it’s safe for someone opting into JSON Schema, to allow JSON Schema syntax.

> * This schema assumes that the root of a design token file is a group as discussed in [Is the root node of a token tree a standard group? #249](https://github.com/design-tokens/community-group/issues/249). Is that assumption correct?

Yes that’s a safe assumption! We should probably have explicit language; we just never did because why would someone have a design system with 1 and only 1 token (keyword “root”).

We allow for `$ref`s now, which means there may be a _non-root_ node where 1 `.json` file may contain 1 token. But it’s important to consider this case a _partial file,_ i.e. invalid in and of itself, never to be consumed on its own. So yes, the root node should always be seen as a “group.”

> * This schema expects 3 components for all color values, though the specification states "the number of components depends on the color space." Should it be made more accurate to the spec or kept like this for now?

Yes we should allow for any number of components. This is more just future-proofing, because even though the currently-supported colorspaces all _happen_ to have 3 components, we plan on expanding these and in the future, CMYK colorspaces would have 4+. Changing it now would just mean not having to change it later.

> * This schema is currently written for the `draft-07` JSON schema version. Is that the version that should be targeted?

I’m not sure, actually! VS Code seems to have pretty decent 2020-12 support. Are there certain features you opted out of that would be helpful? I think I’d need to understand the tradeoffs a little more. 

> * Should the tests I have written for the schema also be included in the PR or just the schema?

That would be great! We use [Vitest](https://vitest.dev) in this repo, so as long as they are convertible to that, that would be a great addition.

> * Does this schema work with the arbitrarily-nested DTCG files issue mentioned in [feat: Add JSON Schema for resolver documents #356](https://github.com/design-tokens/community-group/pull/356)?

I was not able to easily test it because of the absolute `$ref` URLs that point to files that don’t yet exist. I left this comment elsewhere, but could we try relative URLs, at least for testing? I have several documents in [dtcg-examples](github.com/terrazzoapp/dtcg-examples) I could test them on.

pnpm-lock.yaml
```diff
@@ -5,7 +5,6 @@ settings:
   excludeLinksFromLockfile: false
```
drwpow: I think we can discard these changes, since no packages are being adjusted

www/public/schemas/2025.10/format/format.json
```diff
@@ -0,0 +1,105 @@
+{
```
drwpow: Minor thing: let’s move this root-level schema one level up. We are already within the `2025.10` folder, I think it would be nicer to type `.../2025.10/format.json` rather than `.../2025.10/format/format.json`. It’s a nice parallel to the module spec living at `/TR/2025.10/[name]`—you just replace `/TR/` and add `.json`

pnpm-lock.yaml
dgtlntv: Jup those were committed accidentally, my bad!

dgtlntv:
> Ah so we have a pnpm workspace. In the root level there’s no test framework, but inside `www/` there is Vitest configured (if you rebase off latest `main`; I forget when it was merged in). For tests in `public` I’d recommend just adding a new `www/tests` folder they can run from, that way we don’t deploy our tests to the website 😄. They should get picked up automatically, but if not, it’s just modifying Vitest’s [include](https://vitest.dev/config/include.html) option.
> 
> I do think the tests are optional—if you have some that are ready, or they wouldn’t take much time, we’d love to add them! But I don’t want to make this PR blocking. I tested it again and it works great 🙂

Ah right I didn't think of looking in the main branch 😅.
Maybe it would actually make sense to contribute the tests in a follow up PR. I created quite a lot of fixtures (about 4200 lines) and I think it might be easier to review if it is a separate PR? 
Maybe that could also give room for them to be considered more broadly. Not only as something that is being used to test the JSON schema but also as something that can be provided to design token tool makers to test their tools against as discussed in #247.

Also from our conversation in #356 I was not sure wether I should contribute the resolver schema that I had created, since you have already worked on one. Did you want me to contribute that one too?

Issue #348, opened 2025-11-10
First-Class Mode Support with `$modes` Array


caoimghgin:
# First-Class Mode Support with `$modes` Array

**Date:** 2025-11-09
**Proposed By:** Kevin Muldoon
**Target:** Design Tokens Community Group (DTCG) Specification
**Relates To:** DTCG 2025.10, Section 6.2 "Root Tokens in Groups" (PR #298)

---

## TL;DR

This proposal introduces a `$modes` array to address the variant discoverability challenge in PR #298's `$root` pattern. The `$modes` array provides explicit enumeration of mode variants, positional defaults (first element), and forward-compatible parsing—completing the work started in PR #298 by adding the missing standardized discovery mechanism.

---

## Executive Summary

PR #298 (October 2025) introduced `$root` for "base values within groups" with "variants or extensions" as sibling keys. However, **the spec currently lacks a standardized mechanism for parsers to programmatically discover which sibling keys represent variants**. This requires implementers to use heuristics that may become fragile when the spec evolves.

**The solution:** An explicit `$modes` array that:
- ✅ Eliminates parsing ambiguity (variants are explicitly enumerated)
- ✅ Provides simple default semantics (first element is default, replacing `$root`)
- ✅ Maintains backward compatibility (optional, non-breaking)
- ✅ Enables forward-compatible extensibility (new spec fields never conflict with modes)
- ✅ Complements the existing Resolver specification (different use cases)

**Why now:** Given PR #298's recent timing, we have an optimal window to address this implementation challenge before tools ship with divergent approaches.

---

### What PR #298 Introduced

**Section 6.2 "Root Tokens in Groups"** example:

```json
{
  "accent": {
    "$root": { "$type": "color", "$value": "#dd0000" },
    "light": { "$type": "color", "$value": "#ff2222" },
    "dark": { "$type": "color", "$value": "#aa0000" }
  }
}
```

### The Remaining Implementation Challenge

**Critical question:** How does a parser programmatically discover that `light` and `dark` are variants?

The spec currently lacks a standardized approach. Parsers must use heuristics:

```typescript
// Blacklist approach (may become fragile)
const variantKeys = Object.keys(accent).filter(key =>
  key !== '$root' &&
  key !== '$type' &&
  key !== '$description'
  // What about future spec additions?
  // What about custom metadata?
)
```

**Fundamental challenges:**

1. ❌ Cannot reliably enumerate all variants
[...]

PR #364, opened 2025-12-15
fix: use real token types in examples, remove light/dark examples

drwpow:
## Changes

- Replaces placeholder token values with real values
- Removes confusing “light” / “dark” example mentioned in #348 

## How to Review

- See preview link


ilikescience (approved):
Thanks! Let's merge!

PR #365, opened 2025-12-15
Update resolver schema

dgtlntv:
## Changes
As discussed in #360 I am updating the resolver schema with the resolver schema [that I created](https://github.com/dgtlntv/dtcg-json-schema).

## How to Review
- Test this schema with correct and incorrect resolver files and check wether it accurately validates the resolver files.


dgtlntv:
I added a small drive by fix to the format schema so it (and the resolver schema) pass strict mode as discussed in #356.

Issue #383, opened 2026-01-16
Conflicting language in "References and JSON Pointer Integration"


mindplay-dk:
I noticed some conflicting language in the "References and JSON Pointer Integration" part of the spec.

"tools **MAY** also support JSON Pointer notation"

https://github.com/design-tokens/community-group/blob/21dac16da36e70abf8b55d92c985dc8816f21ea1/technical-reports/format/groups.md?plain=1#L511

"Tools **MUST** support JSON Pointer references"

https://github.com/design-tokens/community-group/blob/21dac16da36e70abf8b55d92c985dc8816f21ea1/technical-reports/format/groups.md?plain=1#L534

In addition, the spec says the `{group.token}` syntax is "**maintained for backward compatibility**"

https://github.com/design-tokens/community-group/blob/21dac16da36e70abf8b55d92c985dc8816f21ea1/technical-reports/format/groups.md?plain=1#L511

Yet, the `{group.token}` syntax is defined under the title "**Current** Reference Syntax (**Recommended**)"

https://github.com/design-tokens/community-group/blob/21dac16da36e70abf8b55d92c985dc8816f21ea1/technical-reports/format/groups.md?plain=1#L513

Has the working group decided on a syntax?

Personally, I prefer the simpler syntax, which is immediately easy to parse and implement - whereas JSON pointers require a library.

JSON pointers would let you write valid expressions that reference array elements - is that useful or does it create problems?

JSON pointers would also let you reference reserved schema elements like `$type` and `$name` - is that a "by design" feature or by accident?

Does simpler syntax lack anything, or why were JSON pointers added with a "MUST" requirement?

drwpow:
@mindplay-dk

> That reference would have been much more reliable as e.g. {brand.primary.lightness} - a computed property that could be calculated for any color regardless of how you happened to define that color.

I want to point out, that this has nothing to do with JSON pointer syntax. Aliases have the same limitation—numbers are numbers, and not complex computed properties that translate across color spaces (which, I’ll add, is not even mathematically possible—translation between color spaces [is opinionated](https://bottosson.github.io/posts/gamutclipping/) which includes all components, not limited to lightness).

I think what you may be wanting here is something like [TokenScript](https://docs.tokenscript.dev.gcp.tokens.studio) or [DesignTokens Lang](https://docs.tokenscript.dev.gcp.tokens.studio), which are experimental examples of more complex computation than the current specification currently allows (which are very cool, by the way!).

[...]

PR #387, opened 2026-01-18
JSON schema references fix in non-composite tokens

dgtlntv:
## Changes

This PR fixes a bug in the JSON schema where curly brace references were allowed within sub-values of non-composite tokens (which is not allowed). See #383 for discussion. Only composite tokens explicitly allow curly brace references in sub-values according to the specification [here](https://www.designtokens.org/tr/2025.10/format/#composite-types).

## How to Review
- Review the JSON schema of the non-composite token types and make sure that within them only the `jsonPointerReferenceObject` definition is used but not the `tokenValueReference` definition (as the `tokenValueReference`  allows for both a `curlyBraceReference` as well as a `jsonPointerReferenceObject`).
- Test with a design token file which incorrectly has a curly brace reference in a sub-value of a non-composite token and check wether the schema properly flags it.

### DIFF
--- a/www/public/schemas/2025.10/format/values/color.json
+++ b/www/public/schemas/2025.10/format/values/color.json
@@ -45,7 +45,7 @@
           "maximum": 1
         },
         {
-          "$ref": "../../format.json#/definitions/tokenValueReference"
+          "$ref": "../../format.json#/definitions/jsonPointerReferenceObject"
         }
       ]
     },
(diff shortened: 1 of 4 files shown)


drwpow (approved):
Thank you! I expect with dozens of edge cases like this we may have a few more fixes like these (I’ve probably also introduced some mistakes myself—all part of the process 🙂)

PR #389, opened 2026-01-30
Allow whole array aliasing in JSON schema

dgtlntv:
## Changes

Another small fix to the JSON schema. For the color and strokeStyle token values the JSON schema previously did not allow for the whole array in those schemas to be referenced with a JSON pointer. I.e.

```json
{
  "colors": {
    "$type": "color",
    "base-blue": {
      "$value": {
        "colorSpace": "srgb",
        "components": [0.2, 0.4, 0.9]
      }
    },
    "derived-same-components": {
      "$description": "References the entire components array from base-blue",
      "$value": {
        "colorSpace": "srgb",
        "components": { "$ref": "#/colors/base-blue/$value/components" }
      }
    }
  }
}
```
was not allowed before but from my understanding it should be allowed.

In the color value schema since the components can now be an array or an object I needed to refine the `if` statements to check for the color space and wether the components are an array (so it does not try to apply color space rules on components that are a JSON pointer reference).

## How to Review
Verify if the schema now properly allows referencing whole arrays within color and strokeStyle values. Also verify that that this change did not break any valid uses of color or strokeStyle values.

### DIFF
--- a/www/public/schemas/2025.10/format/values/color.json
+++ b/www/public/schemas/2025.10/format/values/color.json
@@ -34,7 +34,14 @@
     },
     "components": {
       "description": "Array of color components. The number and meaning of components depend on the color space. Each component can be a number or the 'none' keyword.",
-      "type": "array"
+      "oneOf": [
+        {
+          "type": "array"
+        },
+        {
+          "$ref": "../../format.json#/definitions/jsonPointerReferenceObject"
+        }
+      ]
     },
     "alpha": {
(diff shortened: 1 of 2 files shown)


drwpow (approved):
This looks good! I don’t have the ability to test it this week, but I think this would be safe to merge just to test, and we can revert if we find something broke (I’m OK with a little bit of thrash up front—it’s better to try these improvements sooner than later!)

PR #394, opened 2026-02-24
Fix docs internal broken links

julien-deramond:
> [!WARNING]
> This PR includes the content https://github.com/design-tokens/community-group/pull/393 as it's a bit complicated to stack branches when it's a fork. This would need to be rebased after the merge of https://github.com/design-tokens/community-group/pull/393.

## Changes

This PR fixes all the internal broken links found out by https://github.com/design-tokens/community-group/pull/393.

As you can see, the GitHub jobs are now 🟢 

All fixes are handled in separate commits to ease the review.


drwpow (approved):
Thank you! This looks great.

This is not too much to review, I’d rather just merge this PR since this also includes the link checker changes.

I’m fine NOT checking remote links for now. To be honest I‘d rather not ping random servers on every PR and CI run, a) to be kind to their bandwidth but b) as a CI practice if a remote website goes down I don’t want us to be blocked (there’s also a security aspect here too, but if we’re worried about these sites’ security we shouldn’t link to them in the first place). Remote link breaks can just be manually reported/fixed for now!

PR #412, opened 2026-06-02
Bundle JSON schema

dgtlntv:
## Changes

This PR moves JSON schemas from `www/public/schemas/` into a dedicated `schemas` package with modular source files and a bundling step. The build script uses `@hyperjump/json-schema` to bundle split schema files (with `$ref`) into self-contained output files.

As brought up by @drwpow in this [PR discussion](https://github.com/design-tokens/community-group/pull/356) schema store expects bundled schemas if the schema is stored somewhere else. I kept the source as separate schemas +  a bundling step because I think the schema would be easier to maintain this way, especially once the specification grows more.

Having the schema be a separate package (rather than be part of the `www` package) could have other positive effects. It would be easier to use the schema to test other packages. For example the schema could be used as part of a test in the `technical-reports` package to test whether the code snippets inside it are valid design token json. (I would contribute this as a separate PR).

As discussed in [this PR](https://github.com/design-tokens/community-group/pull/356)  the bundled schema would not work in VScode right now. But [the PR](https://github.com/microsoft/vscode-json-languageservice/pull/308#issuecomment-4423217677) for the `vscode-json-languageservice` that is fixing this issue seems to be close to being merged. We could delay this PR until that one lands in the `vscode-json-languageservice`.

In a followup PR I would add tests for the schemas package. These tests would run the schemas against fixtures using both `ajv` and `vscode-json-languageservice` that way we can be sure in the future that the JSON schema works properly with VScode.

After this PR is merged we could proceed with contributing this schema to [schemastore](https://github.com/SchemaStore/schemastore/issues/5290).

## How to Review

* Verify the bundled schema still works properly to validate design token json files (for example with ajv or similar)
* Verify the www package builds properly and serves the bundled json schema
* Verify the build script and package set up aligns with this repos code standards etc.


schemas/scripts/bundle.ts
```diff
@@ -0,0 +1,157 @@
...
+  mkdirSync,
+} from 'node:fs';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import * as prettier from 'prettier';
+import {
+  registerSchema,
+  type SchemaObject,
+} from '@hyperjump/json-schema/draft-07';
+import { bundle } from '@hyperjump/json-schema/bundle';
```
drwpow, line 24: Nice find!

drwpow (approved):
This is great, thank you for taking the time to do this! This had dropped off my list, apologies. Appreciate you taking the time.

I like your choices in regard to a new schemas package—that keeps it clean and debuggable. And library seems like a great choice for bundling.

I don’t have any feedback at this stage, let’s just ship this and adjust if anything comes up. This seems like a great foundation to build off of.

Issue #386, opened 2026-01-18
Design token test fixtures


dgtlntv:
### Problem area

Other

### Problem

When I was working on the JSON schema which has been recently added to this repository I created design token test fixtures to test wether the JSON schema works correctly. 
I was thinking that these test fixtures might not only be valuable to test the JSON schema itself but also as something that could be shared on its own. Design token tool builders could use these fixtures to test their tools too. That way not every tool developer would have to write their own fixtures. Something that was previously mentioned [here](https://github.com/design-tokens/community-group/issues/247#issuecomment-2304377296).

I created a draft of what something like this could look like [here](https://github.com/dgtlntv/community-group/tree/tests/tests).
I tried to take [inspiration from the test suite of json-ld](https://github.com/w3c/json-ld-api/tree/main/tests) in how the fixtures and manifests are set up.

Is that something that could be interesting or should I contribute the test fixtures only as something to test the JSON schema itself?

### Required

- [x] I have searched [existing issues](https://github.com/design-tokens/community-group/issues) and there is not already an open issue on the subject.
- [x] I have read and agree to abide by the [CODE_OF_CONDUCT](https://github.com/design-tokens/community-group/blob/CODE_OF_CONDUCT.md)

PR #413, opened 2026-06-03
Add test suite package

dgtlntv:
(This PR is stacked on top of #412. So the diff also includes the schema package changes from that PR. The new changes in this PR are focused on the test suite package and the schema tests that consume it.)

## Changes                                                                                                                                                                   
This PR introduces a test suite as a separate package. I originally created these fixtures to test the JSON schema while working on it, but I think they may also be useful to third-party design token tool developers. At the moment, every design token tool developer has to write their own fixtures to test things like token parsers. Being able to use official fixtures could make that easier.

For the DTCG, this gives the fixtures a dual purpose: they let us test the JSON schemas we publish, and they provide official test fixtures for the wider community.

This PR also adds tests to the `schemas` package that use the fixtures to test the bundled schemas. These tests currently run with `ajv`. I also added a `vscode-json-languageservice`  test setup so we can verify that the JSON schemas work in VS Code as well. Those tests are currently skipped because they do not pass yet. We need to wait for the upstream fix mentioned in #412.
The test setup is structured so it should be easy to add more JSON schema validators in the future if we want to compare behavior across multiple validators.

The test suite package itself is structured similarly to the [JSON-LD Working Group test suites](https://github.com/w3c/json-ld-api/tree/main/tests). Specifically it uses manifest files instead of relying only on folder structure, so each fixture has an explicit id, expected result, purpose, input path, and feature tags. That should make the fixtures easier to consume programmatically, easier to review, and easier to extend over time without forcing every implementation to infer test meaning from filenames alone.

I know this PR is quite big and probably hard to review. LMK if I should split it up in a certain way.                                                                                

## How to Review

- Run the tests in the schemas package and make sure they pass.
- Look at the test fixtures and make sure they make sense.

### DIFF
--- a/schemas/tests/helpers.ts
+++ b/schemas/tests/helpers.ts
@@ -166,7 +166,7 @@ export interface ValidationResult {
   errorDetails: string;
 }
 
-/** Options for customising test generation. */
+/** Options for customizing test generation. */
 export interface GenerateTestsOptions {
   /**
    * Test case IDs to exclude from the generated suite.
--- a/schemas/tests/helpers.ts
+++ b/schemas/tests/helpers.ts
@@ -1,0 +1,252 @@
+/**
+ * Shared test infrastructure for schema validation tests.
+ *
+ * Provides types, config loading, manifest parsing, and test
+ * generation utilities used by all validator-specific test files.
+ */
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from 'vitest';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const ROOT_DIR = join(__dirname, '..');
+
+// ---------------------------------------------------------------------------
+// Manifest types (mirrors the structure in test-suite)
+// ---------------------------------------------------------------------------
+
+/** A single test case entry in a sub-manifest. */
+export interface TestCase {
+  /** Unique identifier for the test. */
+  id: string;
+  /** Whether the fixture should pass or fail validation. */
+  type: 'PositiveEvaluationTest' | 'NegativeEvaluationTest';
+  /** Human-readable test name. */
+  name: string;
+  /** Explanation of what the test verifies. */
+  purpose: string;
+  /** Path to the fixture file, relative to the sub-manifest directory. */
+  input: string;
+  /** Optional list of features exercised by this test. */
+  features?: string[];
+}
+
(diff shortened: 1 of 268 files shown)


dgtlntv:
This would fix issue #386

schemas/tests/validate-vscode.test.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * Schema validation tests using the VS Code JSON language service.
```
drwpow, line 2: :100::100::100: I can’t believe we’ll get these (well, once the issue mentioned is fixed upstream)!!

drwpow (approved):
These look great! And I see that the `pnpm --recursive run test` already picked up the new tests with the "test" npm script. I’ll wait for this to rebased off `main`, but otherwise LGTM and can’t wait to ship!

schemas/tests/helpers.ts
```diff
@@ -0,0 +1,269 @@
...
+ * Result of validating a single fixture.
+ */
+export interface ValidationResult {
+  /** Whether the fixture passed validation. */
+  isValid: boolean;
+  /** Formatted error/diagnostic details for failure messages. */
+  errorDetails: string;
+}
+
+/** Options for customising test generation. */
```
kaelig: Use US English "customizing" (and then remove "customising" from terminology.txt)

dgtlntv: Done! :)

dgtlntv:
> These look great! And I see that the `pnpm --recursive run test` already picked up the new tests with the "test" npm script. I’ll wait for this to rebased off `main`, but otherwise LGTM and can’t wait to ship!

Rebased it and addressed @kaelig feedback as well! 

Once the fix lands for the VSCode JSON language service I would create a followup PR that un-skips the VSCode tests.

dgtlntv:
@drwpow I updated the dependency for `vscode-json-languageservice` to `6.0.0-next.2` where it now supports the bundled schema so the vscode json languageservice tests can be unskipped (which I changed as well). I checked what version vscode itself is using and it also seems to be using  `6.0.0-next.2` so the bundled schema should now also work properly in vscode.

Which I guess is also a good time to make sure the schema is also added to schema store. It looks like someone already opened a [PR](https://github.com/SchemaStore/schemastore/pull/6207) for it.

drwpow:
That all sounds fine. Thanks for all the hard work on this, and sorry for the delayed review! (This dropped off my list)

Issue #368, opened 2025-12-17
Example 25 uses a not standard unit (%)


Nyan11:
### Problem area

Other

### Problem

Hello,

The example 25 uses a not standard unit:

[Example 25](https://www.w3.org/community/reports/design-tokens/CG-FINAL-format-20251028/#example-group-extension-with-override): Group extension with override

```txt
"width": { "$value": { "value": 100, "unit": "%" } },
```

The % unit is not a standard dimension according to https://www.w3.org/community/reports/design-tokens/CG-FINAL-format-20251028/#dimension

### Required

- [x] I have searched [existing issues](https://github.com/design-tokens/community-group/issues) and there is not already an open issue on the subject.
- [x] I have read and agree to abide by the [CODE_OF_CONDUCT](https://github.com/design-tokens/community-group/blob/CODE_OF_CONDUCT.md)

PR #414, opened 2026-06-03
Validation test for design token JSON examples in the technical report

dgtlntv:
(This PR is stacked on top of #412. So the diff also includes the schema package changes from that PR. The new changes in this PR are focused on the added validation test in the technical report package.)

## Changes

Adds a Vitest-based validation test for fenced `json` and `jsonc` examples in the `technical reports` package.

Previously, examples in the technical reports relied on manual inspection. This test automatically extracts examples from the markdown files and validates them against the bundled DTCG JSON Schemas, making it easier to catch invalid examples and markdown/code-block issues.                                                                                           
                                                                                                                                                                                         
The test currently checks three things:
1. examples parse as valid JSONC
2. blocks tagged as `json` do not contain JSONC-only features such as comments
3. examples with a known `$schema`, or examples detected as DTCG format/resolver examples, validate against the corresponding bundled schema.
                                                                                                                         
Configuration lives in:

```txt
technical-reports/tests/tests.config.json
```

The config defines:
- which markdown files are scanned
- where the bundled schemas are loaded from
- detection rules for examples without $schema
- exclusions for examples that are intentionally invalid/incomplete.
 
### Current Result

Running the new test currently reports 37 failures (which I think almost all of them are valid failures). I wasn't sure whether I should fix those issues in this PR or in a follow up PR.   

## How to Review

Run:
```sh
pnpm --filter @dtcg/tr test
```

Check whether the reported failures are legitimate and whether the test behavior/configuration is appropriate.


dgtlntv:
This would address issue #368

dgtlntv:
Hi @drwpow! Just wanted to check if there’s still interest in this PR. If not, I can close it.

technical-reports/package.json
```diff
@@ -30,8 +32,13 @@
     "validate:resolver": "respec --localhost resolver/index.html --haltonerror --haltonwarn"
   },
   "devDependencies": {
+    "ajv": "^8.17.1",
+    "ajv-formats": "^3.0.1",
     "chokidar": "^4.0.3",
     "chokidar-cli": "^3.0.0",
-    "respec": "^35.6.1"
+    "glob": "^11.0.0",
```
drwpow, line 39: tinyglobby is a newer, faster glob library that comes “for free” in Vite etc. and is now better-maintained than glob. Let’s use that instead

```suggestion
    "tinyglobby": "^0.2.17",
```

technical-reports/package.json
```diff
@@ -2,13 +2,15 @@
...
   "version": "0.0.0",
+  "type": "module",
   "description": "Technical reports for the design tokens specification.",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/design-tokens/community-group.git",
     "directory": "technical-reports"
   },
   "scripts": {
+    "test": "vitest run",
```
drwpow, line 13: 💯

drwpow (approved):
Yes, still interest in this! This all looks great. This is a fantastic start, I have no problems with just running this, and modifying it as we go

